PR #86, opened 2024-01-02
[UI] refactor routing layer

KaradzaJuraj:
## Changes:
 - removed `config/ApplicationRoutes`
 - added `routes` folder with:
   - `PathConstants` -> all available **paths** are saved here, from now on we can access them like `PathConstans.{displays_available_options}`
   - `index` -> all available **routes** are saved here with their corresponding component which is being lazy loaded (loads only when needed instead of loading all of them at the start)
 - refactored the `src/index` (starting point of our app) because it was using `ReactDOM.render(...)` which is deprecated
 - `src/App` is the router now
 - `src/AppLayout` contains the Basic Layout of our application
   - `<Suspense>` is for lazy loading, gives an message while the page is rendering
   - `<Outlet/>` gets changed depending on the page we are rendering at a given moment
 - added `Page404` for unspecified routes
 - fix the BE host endpoint for `history`
 - removed the collapse function since it wasn't working anyways and we have no need for it
 
 #### _Once the PR is approved, the changes on dependencies will be committed (100+ files)._


petar-cvit (approved):
Looks good ❤️

PR #90, opened 2024-01-05
[UI] remove the search bar component

KaradzaJuraj:
The export from the `searchbar` component had no usage in the entire project.
I tested it out on the demo template, and it seemed to work fine.
Nodes were functional as well.


petar-cvit (approved):
Thanks @KaradzaJuraj!

PR #91, opened 2024-01-05
Show all resources deployed by a Module

petar-cvit:
When listing resources by Modules, we should check for all GVKs we can show and that were deployed by a certain Module. This PR refactors how we fetch stuff from the BE.

Also, finally extracted each K8s Kind into a separate React component. Seems much easier to implement a separate view for new components

### DIFF
--- a/cyclops-ui/src/components/k8s-resources/StatefulSet.tsx
+++ b/cyclops-ui/src/components/k8s-resources/StatefulSet.tsx
@@ -1,0 +1,276 @@
+import React, {useEffect, useState} from 'react';
+import {Button, Col, Divider, Row, Table, Typography, Alert, Tag, Tabs, Modal, TabsProps} from 'antd';
+import {useNavigate} from 'react-router';
+import axios from 'axios';
+import {formatPodAge} from "../../utils/pods";
+import {CheckCircleTwoTone, CloseSquareTwoTone, DownloadOutlined} from "@ant-design/icons";
+import ReactAce from "react-ace";
+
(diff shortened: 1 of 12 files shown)


KaradzaJuraj (changes requested):
Most of the comments are repating from file to file, so its actually  not that much.
I still need to review the `module_details` file, which should be renamed to ModuleDetails .

cyclops-ui/src/components/k8s-resources/StatefulSet.tsx
```diff
@@ -0,0 +1,262 @@
...
+    };
+
+    const downloadLogs = (container: string) => {
+        return function () {
+            window.location.href = window.__RUNTIME_CONFIG__.REACT_APP_CYCLOPS_CTRL_HOST + '/resources/pods/' + logsModal.namespace + '/' + logsModal.pod + '/' + container + '/logs/download';
+        }
+    }
+
+    const getTabItems = () => {
+        var items: TabsProps['items'] = []
```
KaradzaJuraj, line 81: use `let` instead of `var`

KaradzaJuraj (changes requested):
Reviewed the ModuleDetails file.

We have warnings on the `<Collapse>` component because it's using `children` instead of `items`, and an error caused by our components not having unique keys.

Fixing these warnings would require a lot of code changes which would be better addressed in a separate PR. This is fine for now

KaradzaJuraj (approved):
LGTM, thanks for cleaning it up ❤️

Issue #87, opened 2024-01-02
Support OCI based Helm


petar-cvit:
**Is your feature request related to a problem? Please describe.**
OCI based registries are not supported when pulling Helm chart dependencies

**Additional context**
Connected to https://github.com/cyclops-ui/cyclops/issues/75

petar-cvit:
OCI charts are now supported for dependency charts.
Root OCI charts are yet to be supported

petar-cvit:
Available from `v0.0.1-alpha.9`

PR #93, opened 2024-01-09
OCI registries support

petar-cvit:
Added support for OCI registries [https://helm.sh/docs/topics/registries/](https://helm.sh/docs/topics/registries/)

Resolves https://github.com/cyclops-ui/cyclops/issues/87


KaradzaJuraj (approved):
Looking good 👍

PR #106, opened 2024-01-22
Suggest versions for a selected template

petar-cvit:
Return possible versions for a selected template, making usage of a template version easier.

Will help prevent wrong (unexistent) template versions.

Support fetching versions for different sources of templates
 - git - branches and tags
 - helm repo - semantic versions
 -  OCI - semantic versions


petar-cvit:
Won't implement since we refactored the template selection flow

PR #112, opened 2024-02-05
🎀 [UI] Prettier plugin

KaradzaJuraj:
## Changes:
- add prettier plugin
- tested it out on `App.tsx` and `index.tsx`


cyclops-ui/src/index.css
petar-cvit, line 1: can we remove this if its empty

Issue #127, opened 2024-02-23
Remove Alert Upon Successful Load of Charts


KaradzaJuraj:
**Is your feature request related to a problem? Please describe.**
When creating a new Module, after successfully loading the chart, an alert pops up on the top of the screen indicating success. This alert is unnecessary and should be removed.

**Describe the solution you'd like**
Remove the success alert when loading templates.

petar-cvit:
Implemented

PR #129, opened 2024-02-24
render form once initial values are loaded

petar-cvit:
When creating a new Module, the UI reports it has loaded the template, but is still loading initial template values. With this PR, a loading Spinner is introduced until both are loaded.


petar-cvit:
Solves https://github.com/cyclops-ui/cyclops/issues/127

KaradzaJuraj (approved):
LGTM 👍

Issue #114, opened 2024-02-06
Add regex validators for the dynamic forms


KaradzaJuraj:
**The issue**
[JSON Schema](https://json-schema.org/understanding-json-schema/reference/regular_expressions) supports more validations that we currently have implemented, namely regex. This will be used for the dynamic forms.

**Describe the solution you'd like**
Add validators in `src/utils/validators/string.tsx` that support matching by pattern in JSON Schema.

PR #144, opened 2024-03-06
[UI] Refactor String Validators and Add Pattern Validator

KaradzaJuraj:
## Related issue
- #114 

## Changes
- add support for `pattern` from `values.schema.json`
- introduce the `Rule` interface imitating the AntDesign rules
- remove all `Promises` and custom implementations for validating fields


cyclops-ctrl/internal/models/templates.go
```diff
@@ -39,6 +39,7 @@ type Field struct {
 	MultipleOf *float64 `json:"multipleOf"`
 
 	// string validation
-	MinLength *int `json:"minLength"`
-	MaxLength *int `json:"maxLength"`
+	MinLength *int   `json:"minLength"`
+	MaxLength *int   `json:"maxLength"`
+	Pattern   string `json:"pattern"`
```
petar-cvit, line 44: Can we make this a pointer as well?

petar-cvit (commented):
Looks good! Can you refactor number validators as well?

PR #169, opened 2024-03-22
[UI] Fix Bug When Setting Errors

KaradzaJuraj:
## Bugfix 
- `setError` was being called with bad data

### Reproducing the bug
1. start your cluster
2. do **NOT** start `cyclops-ctrl`
3. start `cyclops-ui`
4. go to `localhost:3000`

Without this bugfix, you should be getting a red screen!
Note that this is not the only way of showcasing this bug, but is the simplest one.


petar-cvit (approved):
Nice one! Tried looking into it a couple of times but didn't find the cause

PR #182, opened 2024-03-31
Updated link to CRDs manifest

vedrankolka:
The CRDs manifest seems to have moved into the chart directory, but the link for installing CRDs wasn't updated.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,8 @@ You can install minikube and run it using `minikube start`. It will set up a Kub
 > You can install it using the command below:
 >
 > ```zsh
-> kubectl apply -f https://raw.githubusercontent.com/cyclops-ui/cyclops/main/install/chart/crds/cyclops-module.yaml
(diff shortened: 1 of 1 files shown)


petar-cvit:
Hey @vedrankolka, nice catch! I was wondering if there is something I'm going to break by refactoring this into a Helm chart :)

In the meantime, we added another CRD but forgot to update the `CONTRIBUTING.md`. This CRD is also necessary for Cyclops to start properly and needs to be applied to the cluster. The link to the CRD in question is [here](https://raw.githubusercontent.com/cyclops-ui/cyclops/main/install/chart/crds/template-auth-rule.yaml) (it's in the same folder, but another file).

Could you add that one as well? You can add it as a new command or think of some other solution for new contributors to install it using a single command. Again, two commands are completely fine, but if you have something up your sleeve, we'd like to see it. The point is to make it easier for new contributors.

Thanks for the contribution, let us know what you think!

vedrankolka:
I wasn't sure if it was necessary, but now it's added to the command :)

petar-cvit:
Nice, looks good

Issue #155, opened 2024-03-14
Wrong filenames in template dependencies


petar-cvit:
**Describe the bug**
Each template can have dependencies via Helm chart dependencies. Each template has files - either for templates or plain configuration.

When loading dependencies, those files are not named properly and are missing the first letter of the filename. In the example below one of the dependencies is `mysql` and the file is called `ysql/.gitignore` which is obviously missing a letter.
![Screenshot 2024-03-14 at 19 26 06](https://github.com/cyclops-ui/cyclops/assets/72022639/692b415a-7ee7-4362-9c18-f2775beffb60)

**To Reproduce**
1. [Run Cyclops in your Kubernetes cluster and expose it](https://cyclops-ui.com/docs/installation/install/manifest)
2. Send a get request to Cyclops at `/templates` referencing a template that has dependencies
            - you can use [the one we already have](https://github.com/cyclops-ui/templates/blob/filename-dependencies/demo/Chart.yaml). This one has its source on git
            - `curl "http://localhost:3000/api/templates?repo=https://github.com/cyclops-ui/templates&path=demo&commit=filename-dependencies" | jq '.dependencies.[0].files'`
3. You should see that the filename starts with the dependency name, but missing the first letter. The command above is just jq syntax to extract the first dependency from the JSON response and get its files.

**Additional context**
Handler method for the [/templates](https://github.com/cyclops-ui/cyclops/blob/3de848586730e6094d5833829fd73e90a9bbb8bb/cyclops-ctrl/internal/handler/handler.go#L51)

PR #183, opened 2024-03-31
Fixes issue #155 Wrong filenames in template dependencies

vedrankolka:
### Fixes [Wrong filenames in template  dependencies](https://github.com/cyclops-ui/cyclops/issues/155).

`name` is a string containing the filename (including the path).
I don't know if the path is supposed to be included. It looks to me like the original intent was to write:
```
Name: path.Join(parts[1:]...)
```

### Example

For a file `demo/values.yaml`, before the `chart.Name` would be `emo/values.yaml`
Using `name`, it will be `demo/values.yaml`.
Using `path.Join(parts[1:]...)` it will be `values.yaml`.


petar-cvit:
Thanks @vedrankolka

PR #184, opened 2024-04-02
Add CRD for template references

petar-cvit:
To make develoeper experience smoother when creating modules, we added a dropdown when selecting which template you want to use. You can now prepare a templates you will use in the future and just select them in the dropdown instead of writing it each time.

Not to introduce a database to Cyclops, template references were implemented as another CRD which you can apply into the cluster. A single template ref would look like this:
```yaml
apiVersion: cyclops-ui.com/v1alpha1
kind: TemplateStore
metadata:
  name: demo-template
  namespace: cyclops
spec:
  path: demo
  version: main
```


petar-cvit:
kubectl apply -f https://raw.githubusercontent.com/cyclops-ui/cyclops/v0.3.0-rc.2/install/cyclops-install.yaml

PR #215, opened 2024-04-15
💧 sparse checkout chart path

petar-cvit:


petar-cvit:
dont see any improvement with this approach. Closing for now, but might be useful for later

PR #223, opened 2024-04-22
fix(ctrl): removed redis connection

### LINKED ISSUE #195: Remove Redis connection
**Describe the bug**
Previously, Cyclops depended on a Redis storage layer, but it was deprecated. There is still code for connecting to Redis, and it produces a log saying it can't connect to the instance.

That code must be removed to keep the codebase cleaner and not produce misleading logs. This log would pop up on each Cyclops start.

[Here](https://github.com/cyclops-ui/cyclops/blob/e34cfdff30e3989e3a1ac3e7788d6d119eb63fdd/cyclops-ctrl/cmd/main/main.go#L69) is the line that calls the initialization of the Redis client.

It is used for `templatesStorage,` which can also be deprecated and deleted

hanshal101:
Fixes #195 
Closes #195 

## Description
This PR removes the Redis Connection in the `cyclops-ctrl` .
First (after depreciation) we could see the redis connection error in the console.  But after this PR you wont


cyclops-ctrl/internal/handler/handler.go
petar-cvit, line 1: Can you also remove it from line 41 and 42 where its passed to templatesController and moduleController?
```
templatesController := controller.NewTemplatesController(h.templatesStorage, h.templatesRepo, h.k8sClient)
```
Feel free to delete anything that uses `templatesStorage`. If an endpoint relies on that structure, you can remove it. If you are not sure if you should remove it or not, feel free to ask, but anything using it can be deleted

petar-cvit (changes requested):

hanshal101:
Here you go.

petar-cvit (changes requested):
You will also need to delete [this folder](https://github.com/cyclops-ui/cyclops/tree/main/cyclops-ctrl/internal/storage/templates), which implements `templates.Storage` and all occurrences of it.
It's the old way of storing templates, so we can get rid of it. I believe there are still some endpoints using that struct. You can delete those as well

hanshal101:
yup I actually thought of removing this qt first and then resolve all the errors. But I thought this folder might have some exports in other part of the code base. Well will do👍

cyclops-ctrl/internal/controller/templates.go
petar-cvit, line 1: You will need to restore this file since some of the endpoints are still used. Only the ones using `c.templates` can be removed, but others need to be here.

cyclops-ctrl/internal/controller/templates.go
```diff
@@ -1,249 +0,0 @@
...
-	kubernetes *k8sclient.KubernetesClient,
-) *Templates {
-	return &Templates{
-		templatesRepo:    templatesRepo,
-		templates:        templates,
-		kubernetesClient: kubernetes,
-	}
-}
-
-func (c *Templates) StoreConfiguration(ctx *gin.Context) {
```
petar-cvit, line 38: You can delete this one since it depends on c.templates

cyclops-ctrl/internal/controller/templates.go
```diff
@@ -1,249 +0,0 @@
...
-	if err != nil {
-		fmt.Println(err)
-		ctx.Status(http.StatusInternalServerError)
-		return
-	}
-
-	ctx.JSON(http.StatusOK, versions)
-}
-
-func (c *Templates) GetTemplate(ctx *gin.Context) {
```
petar-cvit, line 138: We need to keep this one since it uses only c.templatesRepo and is currently used for fetching templates

cyclops-ctrl/internal/controller/templates.go
petar-cvit: I left some examples below on which methods are used and which are not

petar-cvit (changes requested):

Issue #228, opened 2024-04-24
Sort fields from `values.schema.json`


petar-cvit:
**Is your feature request related to a problem? Please describe.**
Cyclops uses an additional field `order` provided in the `values.schema.json` that allows users to set a custom order of fields in the object in their schemas (e.g. [here](https://github.com/cyclops-ui/templates/blob/ed31c853c3d04b736ab089eeaa36e631ed3b3bd8/demo-extended/values.schema.json#L101))
However, most templates won't have this field set and their children will have a random order. The scope of this issue is to order fields alphabetically if `order` is not provided.

Another possible case is that order is defined for just a subset of fields. Fields with a defined order should come before those without set order.

Example:
```json
{
  "properties": {
    "name": {
      ...
    },
    "scaling": {
      ...
    },
    "general": {
      ...
    },
    "probes": {
      ...
    },
    "networking": {
      ...
    }
    "owner": {
      ...
    },
  },
  "order": [
    "name",
    "general",
    "scaling"
  ]
}

```
Should be rendered in the to following order:
1. name
2. general
3. scaling
4. network
5. owner
6. probes

The first three are defined in the `order`, and the last three are sorted alphabetically. Ordered come before alphabetically sort ones.

**Describe the solution you'd like**
Add alphabetical sorting [here](https://github.com/cyclops-ui/cyclops/blob/3ad123f7cdf4ac2c93b1d64f50e123528a2a0027/cyclops-ctrl/internal/mapper/helm.go#L73)

Sheikh-Abubaker:
Hey @petar-cvit I would like to work on this!!

petar-cvit:
@Sheikh-Abubaker thanks, but @RohanMishra315 already opened a PR for it

Sheikh-Abubaker:
> @Sheikh-Abubaker thanks, but @RohanMishra315 already opened a PR for it

@petar-cvit no worries!

Tanmay-21:
@petar-cvit would like to give this a try.

Tanmay-21:
Hey @petar-cvit, I have raised a PR for the requirement - #261

petar-cvit:
Hey @Tanmay-21. @RohanMishra315 already has a PR for this, so let's wait to hear back from him about this issue

PR #230, opened 2024-04-27
Sorted fields from values.schema.json

RohanMishra315:
Issue: #228

1. Sorted the fields from values.schema.json into the alphabetical form when the order is not provided
2.  When order is provided , the defined fields will come in order first and then the sorted fields.

### DIFF
--- a/cyclops-ctrl/internal/mapper/helm.go
+++ b/cyclops-ctrl/internal/mapper/helm.go
@@ -71,17 +71,45 @@ func HelmSchemaToFields(name string, schema helm.Property, dependencies []*model
 }
 
 func sortFields(fields []models.Field, order []string) []models.Field {
-	ordersMap := make(map[string]int)
+    // If no custom order is provided, sort fields alphabetically by name
+    if len(order) == 0 {
+        // Extract names of fields for sorting
+        fieldNames := make([]string, len(fields))
+        for i, field := range fields {
+            fieldNames[i] = field.Name
+        }
 
-	for i, s := range order {
-		ordersMap[s] = i
-	}
+        // Sort field names alphabetically
+        sort.Strings(fieldNames)
 
-	sort.Slice(fields, func(i, j int) bool {
-		return ordersMap[fields[i].Name] < ordersMap[fields[j].Name]
-	})
+        // Map field names to their indices in the sorted array
+        fieldIndices := make(map[string]int)
+        for i, name := range fieldNames {
+            fieldIndices[name] = i
+        }
 
-	return fields
+        // Use the sorted indices to sort the fields
+        sort.Slice(fields, func(i, j int) bool {
+            return fieldIndices[fields[i].Name] < fieldIndices[fields[j].Name]
+        })
+    } else {
+        // Custom order provided, sort fields based on the order
+        ordersMap := make(map[string]int)
+
+        // Map field names to their indices in the custom order
+        for i, s := range order {
+            ordersMap[s] = i
+        }
+
+        // Sort fields based on their indices in the custom order
+        sort.Slice(fields, func(i, j int) bool {
+            return ordersMap[fields[i].Name] < ordersMap[fields[j].Name]
+        })
+    }
+
+    return fields
+}
+
 }
 
 func mapHelmPropertyTypeToFieldType(property helm.Property) string {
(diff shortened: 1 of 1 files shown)


cyclops-ctrl/internal/mapper/helm.go
```diff
@@ -71,17 +71,45 @@ func HelmSchemaToFields(name string, schema helm.Property, dependencies []*model
...
+
+        // Map field names to their indices in the custom order
+        for i, s := range order {
+            ordersMap[s] = i
+        }
+
+        // Sort fields based on their indices in the custom order
+        sort.Slice(fields, func(i, j int) bool {
+            return ordersMap[fields[i].Name] < ordersMap[fields[j].Name]
+        })
```
petar-cvit, line 107: Your solution works when the order is provided for all fields or not provided at all, but it doesn't work when 3/6 fields have the order defined. You can check the issue for further explanation of such a case.
I tried it with the following template, and I got the order from the picture.
```
path: demo-extended
version: field-alphabetical-sort
```
Screenshot 2024-04-28 at 19 07 10

petar-cvit (changes requested):
Thanks @RohanMishra315 for the PR. I left a comment on the sorting logic

RohanMishra315:
// Separate fields with order and without order
        var orderedFields []models.Field
        var unorderedFields []models.Field

        for _, field := range fields {
            if idx, ok := ordersMap[field.Name]; ok {
                orderedFields = append(orderedFields, models.FieldWithOrder{Field: field, OrderIndex: idx})
            } else {
                unorderedFields = append(unorderedFields, field)
            }
        }

        // Sort fields with order based on the order index
        sort.Slice(orderedFields, func(i, j int) bool {
            return orderedFields[i].OrderIndex < orderedFields[j].OrderIndex
        })

        // Reconstruct the sorted fields slice
        sortedFields := make([]models.Field, 0, len(fields))
        for _, field := range orderedFields {
            sortedFields = append(sortedFields, field.Field)
        }
        sortedFields = append(sortedFields, unorderedFields...)

        fields = sortedFields
    }

    return fields
}
I think this would work  I first separate the fields into two groups: those with a defined order and those without. We then sort the fields with a defined order based on the provided order index. Finally, we reconstruct the sorted fields slice by appending the ordered fields followed by the unordered fields. This approach ensures that fields with a defined order are prioritized and appear before the alphabetically sorted ones, addressing the issue you encountered where only a subset of fields had a defined order.

hanshal101:
Hey can you edit the comment again I think there's some mistake

petar-cvit:
Sounds good to me. The only thing I wouldn't do is introduce a new struct, `FieldWithOrder`. Instead, you can still sort them like this:
```go
sort.Slice(fields, func(i, j int) bool {
	return ordersMap[fields[i].Name] < ordersMap[fields[j].Name]
})
```

petar-cvit:
@RohanMishra315 feel free to update the PR with those changes

cyclops-ctrl/internal/mapper/helm.go
```diff
@@ -71,19 +71,67 @@ func HelmSchemaToFields(name string, schema helm.Property, dependencies []*model
...
+        fieldIndices := make(map[string]int)
+        for i, name := range fieldNames {
+            fieldIndices[name] = i
+        }
+
+        // Use the sorted indices to sort the fields
+        sort.Slice(fields, func(i, j int) bool {
+            return fieldIndices[fields[i].Name] < fieldIndices[fields[j].Name]
+        })
+    } else {
```
petar-cvit, line 95: You dont need this part. The code in the else statement should handle it.

petar-cvit (changes requested):

cyclops-ctrl/internal/mapper/helm.go
```diff
@@ -71,19 +71,45 @@ func HelmSchemaToFields(name string, schema helm.Property, dependencies []*model
...
+        for _, field := range orderedFields {
+            sortedFields = append(sortedFields, field.Field)
+        }
+        sortedFields = append(sortedFields, unorderedFields...)
+
+        fields = sortedFields
+    }
+
+    return fields
 }
```
petar-cvit, line 110: You closed the function and returned after closing it

petar-cvit (changes requested):

petar-cvit (approved):
Thanks @RohanMishra315!

PR #247, opened 2024-05-01
Prometheus handler

### LINKED ISSUE #70: Add metrics to the controller
We should add metrics to track the number of deployed modules, and changes the controller does on the cluster.

Decide which additional metrics should we expose.

Expose metrics to prometheus

hanshal101:
Fixes #70 
Closes #70 

# Description

This PR adds Prometheus Metrics to the `cyclops-ctrl`.

# Method
Here I have use the Prometheus Gauge method. Basically it shows the total number of modules in cyclops.
If a user creates a module and the response is `200` then the number increments by 1. And Decrements if the module is deleted.


petar-cvit (commented):
Seems like there are some changes from your previous PR for adding Deployment logs

PR #248, opened 2024-05-01
feat(backend): add prometheus for cyclops

### LINKED ISSUE #70: Add metrics to the controller

hanshal101:
Fixes #70 
Closes #70 

# Description

This PR adds Prometheus Metrics to the `cyclops-ctrl`.

# Method
Here I have use the Prometheus Gauge method. Basically it shows the total number of modules in cyclops.
If a user creates a module and the response is `200` then the number increments by 1. And Decrements if the module is deleted.


petar-cvit (changes requested):
@hanshal101 great job! I left a few comments, but really excited about this one

cyclops-ctrl/internal/prometheus/handler.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+var moduleQueued = prometheus.NewGauge(prometheus.GaugeOpts{
+	Name:      "Modules",
+	Help:      "No of modules Inc or Dec",
+	Namespace: "cyclops",
+})
+
+func init() {
+	prometheus.MustRegister(moduleQueued)
```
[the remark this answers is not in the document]
hanshal101: should I do `panic(err)` or `fmt.Printf("error : %v", err)`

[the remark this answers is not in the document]
petar-cvit: Can you make this a constructor and return the error to main.go and handle it like [this](https://github.com/cyclops-ui/cyclops/blob/59d5a3489c93bfd16e0abe69eb7dbf9cc83906eb/cyclops-ctrl/cmd/main/main.go#L89)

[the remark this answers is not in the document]
hanshal101: works!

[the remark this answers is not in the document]
hanshal101: Well here I see no error is being return while using the `MustRegister` function:

image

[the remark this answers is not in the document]
petar-cvit: Yeah, `MustRegister` is a wrapper around `Register` which returns an error. `MustRegister` checks for error in `Register` and panics if its not nil.

```go
func (r *Registry) MustRegister(cs ...Collector) {
	for _, c := range cs {
		if err := r.Register(c); err != nil {
			panic(err)
		}
	}
}
```
We can use Register or something else that will allow us to handle errors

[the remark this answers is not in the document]
petar-cvit: https://github.com/prometheus/client_golang/blob/853c5de11e8a0ee0742689859f1642996517b479/prometheus/registry.go#L402

hanshal101:
Hey @petar-cvit I think the app crashed when I rebased the code to main, Also, there were some conflicts with the rebase. Should I open another PR or is there something we can do to solve this error/issue ?

cyclops-ctrl/internal/prometheus/handler.go
```diff
@@ -0,0 +1,37 @@
...
+	Help:      "No of modules Inc or Dec",
+	Namespace: "cyclops",
+})
+
+func init() {
+	if err := prometheus.Register(moduleQueued); err != nil {
+		setupLog.Error(err, "unable to connect prometheus")
+		os.Exit(1)
+	}
+}
```
petar-cvit, line 25: Can we make this a constructor and call it from main.go? You can then pass the setupLog not to duplicate it.

petar-cvit: Also, can you remove os.Exit(1) since not having is not a critical error. We can make this configurable later on

petar-cvit (changes requested):

cyclops-ctrl/internal/prometheus/handler.go
hanshal101: Can you please explain me what exactly I have to do, I'm a bit confused!

petar-cvit: I would like to turn `init` function into a constructor that returns a `Monitor` struct that will have the Prometheus gauge as its property. Then init the gauge and register it to Prom.

```suggestion
type Monitor struct {
       ModulesDeployed prometheus.Gauge
}

func NewMonitor() (Monitor, error) {
        m := Monitor{
             ModulesDeployed: prometheus.NewGauge(prometheus.GaugeOpts{
	              Name:      "modules_deployed",
	              Help:      "No of modules Inc or Dec",
	              Namespace: "cyclops",
              })
        }

	if err := prometheus.Register(m.ModulesDeployed); err != nil {
		setupLog.Error(err, "unable to connect prometheus")
		return Monitor{}, err
	}
	
	return m, nil
}
```
Dont judge, I coded this in a GitHub comment

hanshal101: Oh but how would you call this in the `main` function

petar-cvit: you can do
```go
monitor, err := prometheus.NewMonitor(setupLog)
if err != nil {
  setupLog.Error(err, "failed to create monitor")
}
```

hanshal101: Also how to handle the Inc and the Dec func?

petar-cvit: you can turn those into methods of Monitor
```go
func (m Monitor) IncModule() {
  m.ModulesDeployed.Inc()
}
```

petar-cvit (approved):
Thanks @hanshal101

Issue #236, opened 2024-04-29
Sanitize Input when Saving Templates


KaradzaJuraj:
## Problem

When adding templates to Cyclops, you'll need to enter the `Name`, `Repository URL`, `Path`, and `Version`. If you accidentally include an additional slash (`/`) before or after `Repository URL`, `Path`, or `Version`, Cyclops will still save the template. 

However, this extra slash (`/`) leads to problems later on. When you try to use these templates in the Modules tab, Cyclops won’t be able to fetch them properly. 

### Examples

Adding `/` to the end of repo, or the start of path, or end of path:

```yaml
Repository URL: https://github.com/cyclops-ui/templates/
Path: /demo/
Version: main
```

```json
{
    "message": "Error loading template",
    "description": "could not read 'values.schema.json' file; it should be placed in the repo/path you provided; make sure you provided the correct path"
}
```

Adding `/` to the end of `Version`:

```yaml
Repository URL: https://github.com/cyclops-ui/templates
Path: demo
Version: main/
```

```json
{
    "message":"Error loading template",
    "description":"gzip: invalid header"
}
```

## Solution

Update the `CreateTemplateStore` & `EditTemplatesStore` functions to remove the extra `/` from the start and end of `repo`, `path` and `version` values. The functions can be found in `cyclops-ctrl/internal/controller/templates.go`

devesh-2002:
May I take up this issue?

KaradzaJuraj:
Yeah @devesh-2002, go for it 👍

PR #252, opened 2024-05-05
Sanitize Input while saving Templates

devesh-2002:
Issue #236,
I have updated the functions `CreateTemplateStore` & `EditTemplatesStore` to remove `/` from start and end using `Trim` function.
Please let me know, if there are any changes needed in the PR.

### DIFF
--- a/cyclops-ctrl/internal/controller/templates.go
+++ b/cyclops-ctrl/internal/controller/templates.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/gin-gonic/gin"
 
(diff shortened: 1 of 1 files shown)


petar-cvit (changes requested):
Hey @devesh-2002! Thanks for the PR

I tried it locally, and `repo`, `path` and `version` are sanitized only when calling `GetTemplate`. You are still passing unsanitized input to `k8sTemplateStore := mapper.DTOToTemplateStore(*templateStore)` which maps the request to a Kubernetes CR.
Screenshot 2024-05-05 at 11 05 51

You will need to sanitize the templateStore object as well. With this the validation passed when creating templateStore, but it will fail when you try to create a Module using that template

devesh-2002:
Can you check it now, if it is correct?

petar-cvit (commented):
@devesh-2002  I see you replaced tabs with spaces. Can you revert back to tabs?

devesh-2002:
@petar-cvit I have updated it back to tabs now.

petar-cvit (approved):
Thanks @devesh-2002

PR #259, opened 2024-05-07
Impl(backend): add pvc data fetch api functionality

### LINKED ISSUE #258: Implement resource details for PersistentVolumeClaims

If you want to fetch information about a Deployment form default namespace called `demo-app` you would call:
`/resources?group=apps&version=v1&kind=Deployment&name=demo-app&namespace=default`

The goal of the issue is to support PersistentVolumeClaims on the same endpoint. The endpoint for fetching PVC data would look like this: `/resources?group=&version=v1&kind=PersistentVolumeClaim&name=<name>&namespace=<namespace>`.

In the response, you can add the size of the PVC, access mode, and anything else you would find useful.

Tanmay-21:
Implements #258
Closes #258

## Description 
This PR adds support for fetching PersistentVolumeClaims data on /resource endpoint in `cyclops-ctrl`.

### DIFF
--- a/cyclops-ctrl/internal/cluster/k8sclient/client.go
+++ b/cyclops-ctrl/internal/cluster/k8sclient/client.go
@@ -567,6 +567,11 @@ func (k *KubernetesClient) mapPersistentVolumeClaims(group, version, kind, name,
 		return nil, err
 	}
 
+	storage = 0
+	if persistentvolumeclaim.Spec.Resources != nil && persistentvolumeclaim.Spec.Resources.Requests != nil {
+		storage = persistentvolumeclaim.Spec.Resources.Requests.Storage
+	}
+
 	return &dto.PersistentVolumeClaim{
 		Group:     	 group,
 		Version:   	 version,
@@ -574,7 +579,7 @@ func (k *KubernetesClient) mapPersistentVolumeClaims(group, version, kind, name,
 		Name:      	 name,
 		Namespace: 	 namespace,
 		AccessModes: persistentvolumeclaim.Spec.AccessModes,
-		Size: 		 persistentvolumeclaim.Spec.Resources.Requests.Storage,
+		Size: 		 storage,
 	}, nil
 }
 
--- a/cyclops-ctrl/internal/cluster/k8sclient/client.go
+++ b/cyclops-ctrl/internal/cluster/k8sclient/client.go
@@ -270,6 +270,8 @@ func (k *KubernetesClient) GetResource(group, version, kind, name, namespace str
 		return k.mapPod(group, version, kind, name, namespace)
 	case isConfigMap(group, version, kind):
 		return k.mapConfigMap(group, version, kind, name, namespace)
+	case isPersistentVolumeClaims(group, version, kind):
+		return k.mapPersistentVolumeClaims(group, version, kind, name, namespace)
 	}
 
 	return nil, nil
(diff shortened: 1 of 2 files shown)


cyclops-ctrl/internal/models/dto/k8s.go
```diff
@@ -248,6 +248,48 @@ func (s *StatefulSet) SetDeleted(deleted bool) {
...
 }
 
+type PersistentVolumeClaim struct {
+	Group       string 							`json:"group"`
+	Version     string 							`json:"version"`
+	Kind        string 							`json:"kind"`
+	Name        string 							`json:"name"`
+	Namespace   string 							`json:"namespace"`
+	AccessModes []v1.PersistentVolumeAccessMode `json:"accessmodes"`
+	Size        string							`json:"size"`
```
petar-cvit, line 258: You will need to add
```go
	Deleted   bool   `json:"deleted"`
```
in order for GetDeleted to work

petar-cvit (changes requested):
@Tanmay-21 thanks for the PR. Left one comment for the PVC struct

cyclops-ctrl/internal/models/dto/k8s.go
Tanmay-21: Added!

cyclops-ctrl/internal/cluster/k8sclient/client.go
```diff
@@ -559,6 +561,23 @@ func (k *KubernetesClient) mapConfigMap(group, version, kind, name, namespace st
...
+	}
+
+	return &dto.PersistentVolumeClaim{
+		Group:     	 group,
+		Version:   	 version,
+		Kind:      	 kind,
+		Name:      	 name,
+		Namespace: 	 namespace,
+		AccessModes: persistentvolumeclaim.spec.accessModes,
+		Size: 		 persistentvolumeclaim.spec.resources.requests.storage,
```
petar-cvit, line 577: Can you fix this as well? Also, do we need to checkout for nil pointer dereference if Storage is nil?

Tanmay-21: we need deleted value populated here as false?

Tanmay-21: Also I think storage is a necessary field in PVCs so might never be nil

petar-cvit: > Also I think storage is a necessary field in PVCs so might never be nil

I was thinking the same thing, but would still like to be cautious since we are dealing with pointers. If its nil, you can default to 0

Tanmay-21: Added

cyclops-ctrl/internal/cluster/k8sclient/client.go
```diff
@@ -559,6 +561,28 @@ func (k *KubernetesClient) mapConfigMap(group, version, kind, name, namespace st
...
+func (k *KubernetesClient) mapPersistentVolumeClaims(group, version, kind, name, namespace string) (*dto.PersistentVolumeClaim, error) {
+	persistentvolumeclaim, err := k.clientset.CoreV1().PersistentVolumeClaims(namespace).Get(context.Background(), name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	storage = 0
+	if persistentvolumeclaim.Spec.Resources != nil && persistentvolumeclaim.Spec.Resources.Requests != nil {
+		storage = persistentvolumeclaim.Spec.Resources.Requests.Storage
+	}
```
petar-cvit, line 573: Size is a string, so you can default to an empty string. My bad I said 0 in a previous comment 😬 

Also, you will need to initialize storage with `:=`. Another thing is that you are checking for Resources and Resources. Requests, but you need to check for Requests and Resources.Storage()

Storage is a function instead of a field so you need to call it to retrieve Storage object. Then you need to call String() from storage to get it in a string format

```suggestion
	storage := ""
	if persistentvolumeclaim.Spec.Resources.Requests != nil && persistentvolumeclaim.Spec.Resources.Requests.Storage() != nil {
		storage = persistentvolumeclaim.Spec.Resources.Requests.Storage().String()
	}
```

petar-cvit (changes requested):

petar-cvit (approved):
LGTM

petar-cvit:
Thanks @Tanmay-21!

PR #261, opened 2024-05-07
impl(sort fields): based on order and alphabetic priority

### LINKED ISSUE #228: Sort fields from `values.schema.json`

Tanmay-21:
closes #228

## 📑 Description
Added Functionality to sort values based on first the order provided and then the alphabetic priority of remaining values.
Approach : add index numbers based on order to a map against the value names and base on alphabetic order for remaining values + the size of order so that numbers don't clash and alphabetic order gets higher numbers so lesser priority.

## ✅ Checks

### DIFF
--- a/cyclops-ctrl/internal/mapper/helm.go
+++ b/cyclops-ctrl/internal/mapper/helm.go
@@ -71,10 +71,25 @@ func HelmSchemaToFields(name string, schema helm.Property, dependencies []*model
 }
 
 func sortFields(fields []models.Field, order []string) []models.Field {
+	sort.Slice(fields, func(i, j int) bool {
+		return fields[i].Name < fields[j].Name
+	})
+
 	ordersMap := make(map[string]int)
+	consideredField := make(map[string]bool)
 
 	for i, s := range order {
 		ordersMap[s] = i
+		consideredField[s] = true
+	}
+
+	orderSize := len(order)
+
+	for i, v := range fields {
+		if !consideredField[v.Name] {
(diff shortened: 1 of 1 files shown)


petar-cvit:
Hey @Tanmay-21, this was implemented in #230, so Im closing this PR

PR #262, opened 2024-05-08
feat(backend): handle two modules owning the same resource

### LINKED ISSUE #163: Handle two Modules owning the same resource
Each Module will create its own set of resources. When two Modules create the same resource, one of the Modules will report that it has all the resources it should have, while the other one will just say it's empty and that it misses the resource.

If a new Module is created and it wants to create a resource that is already owned by another Module, it should report on the UI that there is a clash in resources. I wouldn't go with overriding other Modules resource ownership

**Additional context**
Resource ownership is defined on each resource with the label `cyclops.module`. For example, Module `demo` created a Deployment `app,` and the app Deployment will have the label `cyclops.module: demo`

hanshal101:
Closes #163 
Fixes #163 

## 📑 Description
This PR solves the issue of handling two modules having the same resources.

## Solution
Added a check before creating if any value of label `cyclops.Module` matches with the current name of the Module. If return false then return with error.

## ✅ Checks

### DIFF
--- a/cyclops-ctrl/internal/controller/modules.go
+++ b/cyclops-ctrl/internal/controller/modules.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/gin-gonic/gin"
+	v1 "k8s.io/api/apps/v1"
 
 	"github.com/cyclops-ui/cyclops/cyclops-ctrl/api/v1alpha1"
 	"github.com/cyclops-ui/cyclops/cyclops-ctrl/internal/cluster/k8sclient"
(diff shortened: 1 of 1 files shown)


petar-cvit (changes requested):
You should check for the label value in the [Reconcile](https://github.com/cyclops-ui/cyclops/blob/cf71be51793bfebe9cfc96cd6eb3cf2b5881241b/cyclops-ctrl/internal/modulecontroller/module_controller.go#L83) function. It is responsible for deploying all resources from templates. And also this does not only apply to Deployments, but to all K8s services deployed by the controler

PR #278, opened 2024-05-20
ft(cyctl): Delete modules functionality

### LINKED ISSUE #239: `cyctl`: delete Module command
The scope of this issue is to implement the delete method for Modules custom resource. The command should look like:
```
cyctl delete module <module-name>
```

And it should simply call a method similarly to [this](https://github.com/cyclops-ui/cyclops/blob/75394437a7b9caac18964fe279958a3c04354891/cyctl/internal/get/modules.go#L13)

siddhantprateek:
closes #239

## 📑 Description
- Implements module deletion functionality via `cyctl` command.
```bash
cyctl delete module <MODULE_NAME>
```
- Added minimum spacing when no modules are present.

## ✅ Checks

## ℹ Additional context

Suggestion:
- I believe, we need to added additionally functionality of deleting multiple modules 
```bash
cyctl delete module MODULE_1 MODULE_2 MODULE_3 ...
```

### DIFF
--- a/cyctl/internal/get/modules.go
+++ b/cyctl/internal/get/modules.go
@@ -16,16 +16,18 @@ func ListModules(clientset *client.CyclopsV1Alpha1Client) {
 		return
 	}
 
-	longestName := 0
+	longestName := 20 // minimum column width
 	for _, module := range modules {
 		if len(module.Name) > longestName {
 			longestName = len(module.Name)
 		}
 	}
 
-	fmt.Println("NAME" + strings.Repeat(" ", longestName-4) + " AGE")
+	headerSpacing := max(0, longestName-4)
+	fmt.Println("NAME" + strings.Repeat(" ", headerSpacing) + " AGE")
 	for _, module := range modules {
 		age := time.Since(module.CreationTimestamp.Time).Round(time.Second)
-		fmt.Printf("%s"+strings.Repeat(" ", longestName-len(module.Name))+" %s\n", module.Name, age.String())
+		nameSpacing := max(0, longestName-len(module.Name))
+		fmt.Printf("%s"+strings.Repeat(" ", nameSpacing)+" %s\n", module.Name, age.String())
 	}
 }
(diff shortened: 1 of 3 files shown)


hanshal101:
Hey @siddhantprateek thanks for the PR, I think you should make some changes here:

1. Modify the `Delete` function something like this (it will help to delete multiple modules at once):
```
func DeleteModules(clientset *client.CyclopsV1Alpha1Client, moduleName []string) { //rest of code }
```
2. You can use something like this to get the input Module name in an array format:
```
Args:  cobra.MinimumNArgs(2)
```
and
```
deleteCommand, resourceNames := args[0], args[1:]
```
And also make changes in the `DeleteModule` function:
```
if len(moduleName) > 0 {}
else { // return error }
```

Other than this I think it works good !

petar-cvit:
Hey @siddhantprateek thanks for the PR. Will review it today.
I'd say we can have this solution that deletes a single module and then implement it for multiple modules in a different PR.

cyctl/internal/get/modules.go
```diff
@@ -16,16 +16,18 @@ func ListModules(clientset *client.CyclopsV1Alpha1Client) {
...
 	}
 
-	fmt.Println("NAME" + strings.Repeat(" ", longestName-4) + " AGE")
+	headerSpacing := max(0, longestName-4)
+	fmt.Println("NAME" + strings.Repeat(" ", headerSpacing) + " AGE")
 	for _, module := range modules {
 		age := time.Since(module.CreationTimestamp.Time).Round(time.Second)
-		fmt.Printf("%s"+strings.Repeat(" ", longestName-len(module.Name))+" %s\n", module.Name, age.String())
+		nameSpacing := max(0, longestName-len(module.Name))
+		fmt.Printf("%s"+strings.Repeat(" ", nameSpacing)+" %s\n", module.Name, age.String())
```
petar-cvit, line 31: Nice catch

petar-cvit (approved):
Thanks @siddhantprateek! Looks good

siddhantprateek:
I've also added bit of error handling to DeleteModules, i'll create separate PR for multiple modules

PR #280, opened 2024-05-20
Revamp module details page

petar-cvit:
Updates:
- changed Collapses on the ModuleDetails page to be easier to read
- implemented a filter  by resource kind on the ModuleDetails page
- added a more verbose error when form validation fails that contains which fields failed, and with which error
Screenshot 2024-05-20 at 19 20 09


KaradzaJuraj (changes requested):
There is leftover code that is commented out and we should remove it if possible

KaradzaJuraj (approved):

PR #281, opened 2024-05-21
ft(cyclops-ctrl): Added support for fetching Secret resources

### LINKED ISSUE #271: Implement resource details for Secrets

If you want to fetch information about a Deployment form default namespace called demo-app you would call:

The goal of the issue is to support Secrets on the same endpoint. The endpoint for fetching Secret data would look like this: /resources?group=&version=v1&kind=Secret&name=<name>&namespace=<namespace>

You can add all the keys of the Secret in the response but not the values. We don't want to show the values since they are secret.

Refer to this PR on adding new types #259

siddhantprateek:
This resolves
closes #271

### Endpoint:
```
GET /resources?group=&version=v1&kind=Secret&name=<secret-name>&namespace=<namespace>
```

Implementation result
```
GET { HOST }/resources?group=&version=v1&kind=Secret&name=justsecret-2&namespace=cyclops
```
Output Response:
```json
{
    "group": "",
    "version": "v1",
    "kind": "Secret",
    "name": "justsecret-2",
    "namespace": "cyclops",
    "type": "Opaque",
    "data": {
        "KEY": "VkFMVUU="
    },
    "deleted": false
}
```

## ✅ Checks

### DIFF
--- a/cyclops-ctrl/internal/models/dto/k8s.go
+++ b/cyclops-ctrl/internal/models/dto/k8s.go
@@ -249,14 +249,14 @@ func (s *StatefulSet) SetDeleted(deleted bool) {
 }
 
 type PersistentVolumeClaim struct {
-	Group       string 							`json:"group"`
-	Version     string 							`json:"version"`
-	Kind        string 							`json:"kind"`
-	Name        string 							`json:"name"`
-	Namespace   string 							`json:"namespace"`
+	Group       string                          `json:"group"`
+	Version     string                          `json:"version"`
+	Kind        string                          `json:"kind"`
+	Name        string                          `json:"name"`
+	Namespace   string                          `json:"namespace"`
 	AccessModes []v1.PersistentVolumeAccessMode `json:"accessmodes"`
-	Size        string							`json:"size"`
-	Deleted   	bool   							`json:"deleted"`
+	Size        string                          `json:"size"`
+	Deleted     bool                            `json:"deleted"`
 }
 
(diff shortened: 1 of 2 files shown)


cyclops-ctrl/internal/models/dto/k8s.go
```diff
@@ -291,6 +291,57 @@ func (p *PersistentVolumeClaim) SetDeleted(deleted bool) {
...
 }
 
+type Secret struct {
+	Group     string            `json:"group"`
+	Version   string            `json:"version"`
+	Kind      string            `json:"kind"`
+	Name      string            `json:"name"`
+	Namespace string            `json:"namespace"`
+	Type      string            `json:"type"`
+	Data      map[string][]byte `json:"data"`
```
petar-cvit, line 301: Can you send only the keys from the secret without values? I would like to allow users to check which secret keys can they use, but without exposing the secret value

siddhantprateek: Cool, is this the response you are expecting?
```json
{
    "group": "",
    "version": "v1",
    "kind": "Secret",
    "name": "justsecret-2",
    "namespace": "cyclops",
    "type": "Opaque",
    "dataKeys": [
        "KEY"
    ],
    "deleted": false
}
```

petar-cvit: Yeah, looks good

petar-cvit (approved):
Thanks @siddhantprateek

PR #307, opened 2024-05-31
Added Back Button Functionality

### LINKED ISSUE #211: Rollback Page needs a `go back` button
### The Problem
- When on the rollback page of a module, there is no way to return to your module page via the app itself (you can do so with the browser commands only). 

### Wanted Solution
- A button that lets you go back to the module page is needed on the page.

Wasiim105:
closes #211 

## ✅ Checks


petar-cvit (approved):
@Wasiim105 thanks

PR #324, opened 2024-06-10
fix #323: cyctl get nonexistent resource

### LINKED ISSUE #323: cyctl get nonexistent resource
**Describe the bug**
When fecthing modules using `cyctl`, you can specify which modules you want to get.
- `cyctl get modules` fetches all modules
- `cyctl get modules demo-1 demo-2` fetches only specified modules (demo-1 and demo-2)

But if you specify a module that doesn't exist, you get an empty response. When the specified module doesn't exist, `cyctl` should  print an error saying there is no module with the specified name - same as kubectl does:
```sh
kubectl get pods i-do-not-exist
Error from server (NotFound): pods "i-do-not-exist" not found
```

Code for fetching modules is [here](https://github.com/cyclops-ui/cyclops/blob/4aa1a59de4bd60aef422448ed6f740ff8509b218/cyctl/internal/get/modules.go#L44)

thebigbone:
closes #323 

## 📑 Description
Added a boolean check which will prevent the empty response if the specified module name doesn't exist.

## ✅ Checks

### DIFF
--- a/cyctl/internal/get/modules.go
+++ b/cyctl/internal/get/modules.go
@@ -40,7 +40,8 @@ func listModules(clientset *client.CyclopsV1Alpha1Client, moduleNames []string) 
 			longestName = len(module.Name)
 		}
 	}
-
+    
+    foundModule := false
 	filteredModules := modules
 	if len(moduleNames) > 0 {
 		nameSet := make(map[string]struct{}, len(moduleNames))
@@ -51,8 +52,12 @@ func listModules(clientset *client.CyclopsV1Alpha1Client, moduleNames []string) 
 		for _, module := range modules {
 			if _, found := nameSet[module.Name]; found {
 				filteredModules = append(filteredModules, module)
+                foundModule = true
 			}
 		}
+        if !foundModule {
+            fmt.Printf("no module found with name: %v \n", moduleNames)
+        }
 	}
 
 	headerSpacing := max(0, longestName-4)
--- a/cyctl/internal/get/modules.go
+++ b/cyctl/internal/get/modules.go
@@ -40,28 +40,41 @@ func listModules(clientset *client.CyclopsV1Alpha1Client, moduleNames []string) 
 			longestName = len(module.Name)
 		}
 	}
-    
-    foundModule := false
+
 	filteredModules := modules
 	if len(moduleNames) > 0 {
 		nameSet := make(map[string]struct{}, len(moduleNames))
 		for _, name := range moduleNames {
 			nameSet[name] = struct{}{}
 		}
-		filteredModules = []v1alpha1.Module{}
+		foundModules := []v1alpha1.Module{}
+		notFoundModules := []string{}
+
 		for _, module := range modules {
 			if _, found := nameSet[module.Name]; found {
-				filteredModules = append(filteredModules, module)
-                foundModule = true
+				foundModules = append(foundModules, module)
+				delete(nameSet, module.Name)
 			}
 		}
-        if !foundModule {
-            fmt.Printf("no module found with name: %v \n", moduleNames)
-        }
+		for name := range nameSet {
+			notFoundModules = append(notFoundModules, name)
+		}
+		if len(notFoundModules) > 0 {
+			for _, name := range notFoundModules {
+				fmt.Printf("no module found with name: %s\n", name)
+			}
+		}
+		filteredModules = foundModules
+
 	}
 
 	headerSpacing := max(0, longestName-4)
-	fmt.Println("NAME" + strings.Repeat(" ", headerSpacing) + " AGE")
+	output := ""
+	if len(filteredModules) > 0 {
+		output += "NAME" + strings.Repeat(" ", headerSpacing) + " AGE\n"
+	}
+
+	fmt.Print(output)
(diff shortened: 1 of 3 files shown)


petar-cvit (changes requested):
Hey @thebigbone, thanks for the PR!

I tried your solution and found that if you specify multiple modules that do not exist, it will print all module names in a single line:
```
./cyctl get modules not-exists another-not-exists
no module found with name: [not-exists another-not-exists]
NAME                 AGE
```
Can we print a line for each Module that does not exist?

Also, if none of the speficied Modules exist, we shouldn't print `NAME AGE` columns since they will be empty.

Another thing is that if you specify one Module that exists and another that doesn't, cyctl wont print that one Module does not exist.
```
./cyctl get modules not-exists exists
NAME                 AGE
exists               4m49s
```

petar-cvit:
To cover all cases:
- specify all that do not exist
```
cyctl get modules not-exists another-not-exists
  
no module found with name: not-exists
no module found with name: another-not-exists
```

 - specify one that does exist and one that doesn't
  ```
  cyctl get modules not-exists exists

  no module found with name: not-exists
  NAME                 AGE
  exists                 28s
  ```

thebigbone:
You're right. My approach was not covering all the possible cases. I have modified the implementation. Thanks for the feedback!

cyctl/internal/get/modules.go
```diff
@@ -47,16 +47,34 @@ func listModules(clientset *client.CyclopsV1Alpha1Client, moduleNames []string)
 		for _, name := range moduleNames {
 			nameSet[name] = struct{}{}
 		}
-		filteredModules = []v1alpha1.Module{}
+		foundModules := []v1alpha1.Module{}
+		notFoundModules := []string{}
```
petar-cvit, line 51: Could you initialize these with `make`?
```suggestion
		foundModules := make([]v1alpha1.Module, 0)
		notFoundModules := make([]string, 0)
```

petar-cvit (commented):
Thanks @thebigbone! Remebered two more things while you are at it:
- Could you also exit with code 1 if there are any not found modules? You can call `os.Exit(1)`
- Also, could you add the same logic to `template_store.go` and `template_auth_rules.go` from the same folder?

thebigbone:
> * Could you also exit with code 1 if there are any not found modules? You can call `os.Exit(1)`
Wouldn't his approach prevent it from showing the existing modules if there are multiple modules specified in the argument, and one of them might have the existing module

thebigbone:
> @thebigbone I would still keep the same output (showing both existing and non existing modules), but would exit with 1 `if len(notFoundModules) > 0`. You would just need to extract `notFoundModules` from the scope of the if statement.
> 
> I see this useful when running cyctl in a script. You can call `cyctl get modules <name>` to see if a module exists by checking the exit status

Got it! I have added the same logic to `template_store.go` and `template_auth_rules.go`.

cyctl/internal/get/modules.go
thebigbone: Done. Also added the exit status.

cyctl/internal/get/template_store.go
```diff
@@ -21,6 +21,7 @@ var (
  cyctl get templates`
 )
 
+// listTemplate retrieves and displays a list of templatestore from the Cyclops API.
```
petar-cvit, line 24: can you remove the duplicated comment

cyctl/internal/get/template_auth_rules.go
```diff
@@ -21,6 +21,7 @@ var (
  cyctl get templateauthrule`
 )
 
+// listTemplateAuthRules retrieves and displays a list of templateauthrules from the Cyclops API.
```
petar-cvit, line 24: can you remove the duplicated comment

cyctl/internal/get/modules.go
```diff
@@ -47,16 +48,36 @@ func listModules(clientset *client.CyclopsV1Alpha1Client, moduleNames []string)
...
+
 	}
 
 	headerSpacing := max(0, longestName-4)
-	fmt.Println("NAME" + strings.Repeat(" ", headerSpacing) + " AGE")
+	output := ""
+	if len(filteredModules) > 0 {
+		output += "NAME" + strings.Repeat(" ", headerSpacing) + " AGE\n"
+	} else {
+		os.Exit(1)
```
petar-cvit, line 77: can you add this check to the end of the function so it prints out the existing modules (if any are found) and then exists with 1?

The output should still look like this:
```
cyctl get modules not-exists exists

no module found with name: not-exists
NAME                 AGE
exists                 28s
```
but should exit with 1

cyctl/internal/get/template_auth_rules.go
petar-cvit, line 1: Can you add exit status logic here as well?

cyctl/internal/get/template_store.go
petar-cvit, line 1: Can you add exit status logic here as well?

petar-cvit (changes requested):

thebigbone:
I have made the changes.

petar-cvit (approved):
@thebigbone thanks! 🧡

PR #329, opened 2024-06-11
add serve command

### LINKED ISSUE #326: `cyctl` expose UI command
## Feature Request
`cyctl` should be able to expose the UI for easy access.

## Solution
Implement a command that will expose the UI to [localhost:3000](http://localhost:3000/) → `cyctl serve`, which should have the same functionality as this command:
```bash
kubectl port-forward svc/cyclops-ui 3000:3000 -n cyclops
```

thebigbone:
closes #326 

## 📑 Description
Command `serve` for port forwarding

## ✅ Checks
- [] I have updated the documentation as required


petar-cvit (changes requested):
Hey @thebigbone, I wouldnt use kubectl from os.exec since that would add a dependency to the cli.

You can check out "k8s.io/client-go/tools/portforward" to do it from golang

thebigbone:
> Hey @thebigbone, I wouldnt use kubectl from os.exec since that would add a dependency to the cli.
> 
> You can check out "k8s.io/client-go/tools/portforward" to do it from golang

Hey! I was able to do the port forwarding for Pods but for Services, I am having a hard time. Here's the program for Pods: https://gist.github.com/thebigbone/862a05714867c9f9259e2439f5886f2b

thebigbone:
I think it's most likely due to the fact that pods have a `/portforward` path in the API but services do not.

image

petar-cvit:
@thebigbone  Oh I get it. You can then directly port-forward one of the pods of `cyclops-ui`. You will just check which pods exist and pick any running pod to port-forward

thebigbone:
> @thebigbone Oh I get it. You can then directly port-forward one of the pods of `cyclops-ui`. You will just check which pods exist and pick any running pod to port-forward

Oh, I thought of that before but wouldn't that drop the connection if the pod terminates for whatever reason (because pods are short lived?)

petar-cvit:
Well, pods shouldn't terminate just like that. I mean, they should be able to terminate without any disruption to the app,  but in this case, it's good enough. If you need to do something quick on your cyclops instance, you can just port forward the pod, do your thing, and close the port forward. On the other hand, if you need a stable connection to cyclops outside of the cluster for a longer period of time, you will need to set up an ingress to expose it.

TL;DR port-forwards are short-lived so having a port-forward rely directly on the pod is completely fine

petar-cvit:
`cyclops-ui` pods are listening on port 80. You can try that

thebigbone:
I have made some changes, now it will take random pod from the service `cyclops`.

thebigbone:
@petar-cvit can you review it?

cyctl/cmd/serve.go
```diff
@@ -0,0 +1,113 @@
...
+
+	config, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		log.Fatal(err)
+	}
```
petar-cvit, line 52: Can you instantiate clientset the same way it was done with [Moduleset](https://github.com/cyclops-ui/cyclops/blob/5fcddebac7e3d9fb93509b6844679c77aa914dfa/cyctl/internal/kubeconfig/kubeconfig.go#L13). You can then just pass it to `startPortForwarding` like it was done [here](https://github.com/cyclops-ui/cyclops/blob/f0d052ed8b735e0d8a418d88d86380342e7d3180/cyctl/internal/get/modules.go#L98)

cyctl/cmd/serve.go
```diff
@@ -0,0 +1,113 @@
...
+	podIndex := rand.Intn(len(podList.Items))
+	podName := podList.Items[podIndex].Name
+
+	roundTripper, upgrader, err := spdy.RoundTripperFor(config)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	path := fmt.Sprintf("/api/v1/namespaces/%s/pods/%s/portforward", "cyclops", podName)
+	hostIP := strings.TrimLeft(config.Host, "htps:/")
```
petar-cvit, line 79: Why are you trimming `htps:/`?

cyctl/cmd/serve.go
```diff
@@ -0,0 +1,113 @@
...
+	serverURL := url.URL{Scheme: "https", Path: path, Host: hostIP}
+
+	dialer := spdy.NewDialer(upgrader, &http.Client{Transport: roundTripper}, http.MethodPost, &serverURL)
+
+	ports := []string{"3000:80"}
```
petar-cvit, line 84: Could we make this a flag? I realize now that users might want to configure that themselves, but we can default to 3000

petar-cvit (changes requested):
Hey @thebigbone, tried it, and it looks good. Left some comments on how to refactor it, but overall it's what we wanted

petar-cvit:
Also, could you run `go mod tidy` in the cyctl? I got some errors when building

cyctl/cmd/serve.go
thebigbone: Like a global variable as the type `*kubernetes.Clientset` and then pass it to the function?

cyctl/cmd/serve.go
thebigbone: Because it already adds `https` when I try to establish the connection:

image

cyctl/cmd/serve.go
petar-cvit: Yeah, same thing as the Moduleset

cyctl/cmd/serve.go
petar-cvit: Is it added because of this line?
```
url.URL{Scheme: "https", Path: path, Host: hostIP}
```

You are trimming `htps:/` which misses a `t`

thebigbone: It handles the duplicate values

cyctl/cmd/serve.go
thebigbone: I was thinking of importing `kubeconfig` but it has Moduleset with a different type

cyctl/cmd/serve.go
```diff
@@ -0,0 +1,120 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/portforward"
+	"k8s.io/client-go/transport/spdy"
+	"k8s.io/client-go/util/homedir"
+)
+
+var Clientset *kubernetes.Clientset
```
petar-cvit, line 26: Can you define and initialize this [here](https://github.com/cyclops-ui/cyclops/blob/5fcddebac7e3d9fb93509b6844679c77aa914dfa/cyctl/internal/kubeconfig/kubeconfig.go#L35-L34)?

That way, you don't have to repeat this:
```go
	var kubeconfig *string
	if home := homedir.HomeDir(); home != "" {
		kubeconfig = flag.String("kubeconfig", filepath.Join(home, ".kube", "config"), "(optional) absolute path to the kubeconfig file")
	} else {
		kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file")
	}
	flag.Parse()

	config, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
	if err != nil {
		log.Fatal(err)
	}

	clientset, err = kubernetes.NewForConfig(config)
	if err != nil {
		log.Fatal(err)
	}
```

cyctl/cmd/serve.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	if len(podList.Items) == 0 {
+		log.Fatal("no pods found for service cyclops-ui in namespace cyclops")
+	}
+
+	rand.Seed(time.Now().UnixNano())
+	podIndex := rand.Intn(len(podList.Items))
+	podName := podList.Items[podIndex].Name
```
petar-cvit, line 77: You can just take the first pod from the list, if there are any

cyctl/cmd/serve.go
```diff
@@ -0,0 +1,120 @@
...
+	serverURL := url.URL{Scheme: "https", Path: path, Host: hostIP}
+
+	dialer := spdy.NewDialer(upgrader, &http.Client{Transport: roundTripper}, http.MethodPost, &serverURL)
+
+	ports := []string{fmt.Sprintf("%d:80", localPort)}
+
+	stopChan, readyChan := make(chan struct{}, 1), make(chan struct{}, 1)
+	out, errOut := new(bytes.Buffer), new(bytes.Buffer)
+
+	forwarder, err := portforward.New(dialer, ports, stopChan, readyChan, out, errOut)
```
petar-cvit, line 95: I think you can just provide os.Stdout and os.Stderr instead of self-created buffers

```suggestion
	forwarder, err := portforward.New(dialer, ports, stopChan, readyChan, os.Stdout, os.Stderr)
```

petar-cvit (changes requested):

cyctl/cmd/serve.go
thebigbone: Using `Stdout` and `Stderr` causes the channel to exit:
```go
        go func() {
                for range readyChan {
                }
                if os.Stdout != nil {
                        log.Fatal("stdout: ", os.Stdout)
                } else if os.Stderr != nil {
                        fmt.Println("stderr: ", os.Stderr)
                }
        }()
```
```shell
Forwarding from 127.0.0.1:3002 -> 80
Forwarding from [::1]:3002 -> 80
2024/06/17 06:34:14 stdout: &{0xc00009a0c0}
exit status 1
```

Creating buffers is needed in this case

cyctl/internal/kubeconfig/kubeconfig.go
```diff
@@ -21,12 +25,18 @@ func GetKubeConfig() func(cmd *cobra.Command, args []string) {
...
 			panic(err.Error())
 		}
 
-		Moduleset, err = client.NewForConfig(config)
+		Moduleset, err = client.NewForConfig(Config)
+		if err != nil {
+			panic(err.Error())
+		}
+
+		Clientset, err = client.NewForConfig(Config)
```
petar-cvit, line 39: You need to use `kubernetes` package. The `client` package does not implement operations over other k8s resources
`Clientset, err = kubernetes.NewForConfig(Config)`

petar-cvit (changes requested):

cyctl/internal/kubeconfig/kubeconfig.go
thebigbone: Sorry about that. I was running the cluster on a different machine and I forgot to change that

cyctl/cmd/serve.go
thebigbone: Removed that part

cyctl/cmd/serve.go
```diff
@@ -0,0 +1,87 @@
...
+	hostIP := strings.TrimLeft(kubeconfig.Config.Host, "htps:/")
+	serverURL := url.URL{Scheme: "https", Path: path, Host: hostIP}
+
+	dialer := spdy.NewDialer(upgrader, &http.Client{Transport: roundTripper}, http.MethodPost, &serverURL)
+
+	ports := []string{fmt.Sprintf("%d:80", localPort)}
+
+	stopChan, readyChan := make(chan struct{}, 1), make(chan struct{}, 1)
+
+	forwarder, err := portforward.New(dialer, ports, stopChan, readyChan, os.Stout, os.Stderr)
```
petar-cvit, line 67:
```suggestion
	forwarder, err := portforward.New(dialer, ports, stopChan, readyChan, os.Stdout, os.Stderr)
```

cyctl/cmd/serve.go
```diff
@@ -0,0 +1,87 @@
...
+	go func() {
+		for range readyChan {
+		}
+	}()
+
+	if err = forwarder.ForwardPorts(); err != nil {
+		log.Fatal(err)
+	}
+
+	fmt.Println("Port forwarding stopped")
```
petar-cvit, line 81: I dont think you need the for loop over the ready chan. Can we have something like this:
```suggestion
	go func() {
		if err = forwarder.ForwardPorts(); err != nil {
			log.Fatal(err)
		}
	}()

	<-readyChan
	fmt.Println("Ready to accept traffic")

	<-stopChan
	fmt.Println("Port forwarding stopped")
```

You just print that port forwarding is ready once you get the message on the `readyChan`

petar-cvit (changes requested):

petar-cvit (approved):
Thanks @thebigbone! 🧡

PR #330, opened 2024-06-11
feat(ui): add component for Secret

### LINKED ISSUE #297: Implement UI component for Secrets
Some K8s components are supported to have a custom UI component, and those can be found [here](https://github.com/cyclops-ui/cyclops/tree/main/cyclops-ui/src/components/k8s-resources). The scope of this issue is to support `Secrets`. The new `Secrets` component should be similar to other components we already have, so feel free to check the existing ones out. You can fetch the data about a secret by calling the backend on

```
/api/resources?group=&version=v1&kind=Secret&name=<name>&namespace=<namespace>
```

To be able to test this, you will need to use a template that supports secrets. You should add the following template and create a Module from it:
```yaml
path: redis
version: test-secrets-ui
```

doncicuto:
closes #297 

## 📑 Description

This PR adds a component for displaying Secrets

## ✅ Checks

## ℹ Additional context

Here's a screenshot of the new component in action. I think that only type and dataKeys are the relevant information to be displayed.

image

### DIFF
--- a/cyclops-ui/src/components/k8s-resources/Secret.tsx
+++ b/cyclops-ui/src/components/k8s-resources/Secret.tsx
@@ -10,13 +10,13 @@ interface Props {
 
 interface secret {
   type: string;
-  dataKeys: string;
+  dataKeys: string[];
 }
 
 const Secret = ({ name, namespace }: Props) => {
   const [secret, setSecret] = useState<secret>({
     type: "",
-    dataKeys: "",
+    dataKeys: [],
   });
   const [error, setError] = useState({
     message: "",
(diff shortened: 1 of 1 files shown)


petar-cvit:
@doncicuto  Could we do something like this so it's easier to check the keys in a secret? You can show the type of secret outside the table.

Screenshot 2024-06-11 at 19 10 16

Values can be just `****`, but with this its more similar to how we show a configmap so it might be easier for users

doncicuto:
Of course @petar-cvit I'll prepare these changes 

Do you want me to add a button next to the key to reveal the actual value after decoding the base64? In this case I'd need to modify the API response

petar-cvit:
@doncicuto you can skip that for now. It would definitely be useful, but I don't want to expose cluster secrets from Cyclops since different users might have different RBAC rules and systems for handling secrets

doncicuto:
Cool @petar-cvit I'll prepare the changes according to your needs, thanks for your feedback

doncicuto:
Hi @petar-cvit I've added the changes that you proposed. Here's a screenshot:

image

Ready for review then

petar-cvit (approved):
Thanks @doncicuto!

PR #349, opened 2024-06-20
fix:implement resource details for Jobs

### LINKED ISSUE #339: Implement resource details for Jobs
Is your feature request related to a problem? Please describe.

If you want to fetch information about a Deployment form default namespace called demo-app you would call:
`/resources?group=apps&version=v1&kind=Deployment&name=demo-app&namespace=default`

The goal of the issue is to support **Jobs** on the same endpoint. The endpoint for fetching Jobs data would look like this:
```
/resources?group=batch&version=v1&kind=Job&name=<name>&namespace=<namespace>
```

You can add data about pods a Job is managing and its status similar to what we are doing with [Deployments](https://github.com/cyclops-ui/cyclops/blob/19726aed0828c999fb17e84100ebea03490666ee/cyclops-ctrl/internal/cluster/k8sclient/client.go#L443).

Refer to this PR on adding new types https://github.com/cyclops-ui/cyclops/pull/259

To be able to test this, you will need to use a template that supports jobs. You should add the following template and create a Module from it:
```
path: demo-job
version: test-job-endpoint
```

patelajay745:
closes #339 

## 📑 Description

created new endpoint : /resources?group=batch&version=v1&kind=Job&name=<name>&namespace=<namespace>

It will response all details of job, start time , end time and details of pod related to it.

## ✅ Checks
- [X] I have updated the documentation as required
- [X] I have performed a self-review of my code

## ℹ Additional context

I have tested code. Screenshot for reference.

Screenshot 2024-06-19 at 9 23 12 PM

### DIFF
--- a/cyclops-ctrl/internal/cluster/k8sclient/client.go
+++ b/cyclops-ctrl/internal/cluster/k8sclient/client.go
@@ -288,6 +288,8 @@ func (k *KubernetesClient) GetResource(group, version, kind, name, namespace str
 		return k.mapSecret(group, version, kind, name, namespace)
 	case isCronJob(group, version, kind):
 		return k.mapCronJob(group, version, kind, name, namespace)
+	case isJob(group, version, kind):
+		return k.mapJob(group, version, kind, name, namespace)
 	}
 
 	return nil, nil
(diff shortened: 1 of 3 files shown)


petar-cvit (approved):
Thanks @patelajay745 🧡

PR #357, opened 2024-06-21
Fix#351 UI representation cronjobs

samarsrivastav:
closes #351 
i have added the cronjob.tsx file  please review it and letme know any changes required.

PR #358, opened 2024-06-21
Fix#351 CronJob added in moduleDetails.tsx

### LINKED ISSUE #351: UI representation for `CronJobs`
### Problem
Some K8s components are supported to have a custom UI component, and those can be found [here](https://github.com/cyclops-ui/cyclops/tree/main/cyclops-ui/src/components/k8s-resources). The scope of this issue is to support `CronJobs`. 

The new `CronJob` component should be similar to other components we already have, so feel free to check the existing ones out. You can fetch the data about a `CronJob` by calling the backend on

```bash
/api/resources?group=batch&version=v1&kind=CronJob&name=&namespace=
```

### Solution
Reference this PR for BE implementation: https://github.com/cyclops-ui/cyclops/pull/346
You can render a similar component to the one we have for [Deployments](https://github.com/cyclops-ui/cyclops/blob/main/cyclops-ui/src/components/k8s-resources/Deployment.tsx).

samarsrivastav:
closes #351 
i have added the cronjob case in the moduledetails.tsx file
 please review it and letme know any changes required.

### DIFF
--- a/cyclops-ui/src/components/pages/ModuleDetails.tsx
+++ b/cyclops-ui/src/components/pages/ModuleDetails.tsx
@@ -378,6 +378,11 @@ const ModuleDetails = () => {
           <Deployment name={resource.name} namespace={resource.namespace} />
         );
         break;
+      case "CronJobs":
+        resourceDetails = (
+          <CronJobs name={resource.name} namespace={resource.namespace} />
+        );
+        break;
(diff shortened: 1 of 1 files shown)


petar-cvit:
Hey @samarsrivastav, thanks for the PR! Could you add this change to #357 so we have it in the same PR?

PR #379, opened 2024-06-28
fix:(#371) Device Port if default then not set is displayed

### LINKED ISSUE #371: Default Service Port Name
### Problem

It is legitimate for a service port to have no value in its name. We should implement a default so that it is not empty when deployed to Cyclops.
Screenshot 2024-06-23 at 15 53 08

### Solution

When a service port has no name, the default display should be `<not set>` You can test it out with the `redis-demo` template

tanbirali:
closes #371

## 📑 Description
checks if namespace is default if default then not set is displayed 
## ✅ Checks

### DIFF
--- a/cyclops-ui/src/components/k8s-resources/Service.tsx
+++ b/cyclops-ui/src/components/k8s-resources/Service.tsx
@@ -80,7 +80,13 @@ const Service = ({ name, namespace }: Props) => {
               title="Name"
               dataIndex="name"
               key="name"
-              render={(text) => (text ? text : "not set")}
+              render={(text) =>
+                text ? (
+                  text
+                ) : (
+                  <span style={{ color: "#A0A0A0" }}>{"<not set>"}</span>
+                )
+              }
             />
             <Table.Column title="Protocol" dataIndex="protocol" />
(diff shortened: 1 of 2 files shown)


petar-cvit (changes requested):
Hey @tanbirali, you are changing the resource namespace here, but you should default to `not set` if the port name is missing in the table [here](https://github.com/cyclops-ui/cyclops/blob/11185d7871fc0957ffab78bfd5f54d783fbb5fc3/cyclops-ui/src/components/k8s-resources/Service.tsx#L79).

tanbirali:
My bad I got it wrong, will revert it and make the necessary changes

tanbirali:
Have a look now, if this is what you wanted 




Screenshot (520)

petar-cvit (commented):
@tanbirali Can we also set the `not set` to be gray so its clear that its not provided? [Here](https://github.com/cyclops-ui/cyclops/blob/504c439dbb6cd72207c4b34d6391a7ebae098d28/cyclops-ui/src/components/pages/TemplateStore/TemplateStore.tsx#L211) is an example of how we did it on another place

petar-cvit:
Closing this one, since #384 is merged

Issue #374, opened 2024-06-25
Namespace filter in the Nodes tab


KaradzaJuraj:
### Problem
We want to be able to view resources deployed only in specific namespaces.

![Screenshot 2024-06-23 at 14 59 11 (1)](https://github.com/cyclops-ui/cyclops/assets/101581637/fd964c16-acca-4209-9e77-229bc00d78fe)

### Solution
Implement filters to display resources per namespace (all / default / cyclops …) similar to how `Name` has a filter.

m2lusen:
Hi, I would be interested in contributing to this.

However, when I started to try and start a branch of the code I encountered the following error when trying to start cyclops-ui.

"warning ..\..\..\..\..\package.json: No license field
$ REACT_APP_CYCLOPS_CTRL_HOST=http://localhost:8888 node_modules/.bin/react-scripts start
'REACT_APP_CYCLOPS_CTRL_HOST' is not recognized as an internal or external command,
operable program or batch file.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command."

Any help would be appreciated

petar-cvit:
Hey @m2lusen, sure thing!

Are you working on windows? If so, can you update the [start script](https://github.com/cyclops-ui/cyclops/blob/main/cyclops-ui/package.json#L48) 
```
"start": "set REACT_APP_CYCLOPS_CTRL_HOST=http://localhost:8888 node_modules/.bin/react-scripts start",
```
to `package.json` and try `yarn start` again. I'll find a way for it to work out of the box.

petar-cvit:
Alternatively, remove the env variable completely from package.json like below
```
"start": "node_modules/.bin/react-scripts start",
```
Set the env var in your cmd (`set REACT_APP_CYCLOPS_CTRL_HOST=http://localhost:8888`), and then run `yarn start` from the same cmd window

Let me know if it works

m2lusen:
Great. Ive got it working now, I had to remove the env variable from package.json, it still wasn't working after updating the start script initially.

Thank you for the help

PR #383, opened 2024-06-28
Namespace filter in the Nodes tab #374

m2lusen:
closes #

## 📑 Description
Added the function getColumnFilterProps, based on getColumnSearchProps. getColumnFilterProps instead of a search bar creates checkboxes for each of the values (namespace in this case). Using handleSearch the selected values are then filtered, to display only the selected values.

getColumnFilterProps was added to namespace

## ✅ Checks
- [ y] I have updated the documentation as required
- [ y] I have performed a self-review of my code

## ℹ Additional context
NA

### DIFF
--- a/cyclops-ui/src/components/pages/NodeDetails.tsx
+++ b/cyclops-ui/src/components/pages/NodeDetails.tsx
@@ -2,6 +2,7 @@ import React, { useEffect, useRef, useState } from "react";
 import {
   Alert,
   Button,
+  Checkbox,
   Card,
   Col,
   Divider,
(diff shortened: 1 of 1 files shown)


kubectl
petar-cvit, line 1: Hey @m2lusen, can you remove this file?

cyclops-ui/src/components/pages/NodeDetails.tsx
```diff
@@ -190,6 +191,97 @@ const NodeDetails = () => {
...
+      );
+
+      return (
+        <div
+          style={{
+            padding: 8,
+            maxHeight: 200,
+            overflowY: "auto",
+            margin: "8px 0",
+            border: "1px solid #ccc",
```
petar-cvit, line 215: Can we remove the border?

cyclops-ui/src/components/pages/NodeDetails.tsx
```diff
@@ -190,6 +191,97 @@ const NodeDetails = () => {
...
+              }}
+            >
+              Close
+            </Button>
+          </Space>
+        </div>
+      );
+    },
+    filterIcon: (filtered: boolean) => (
+      <SearchOutlined style={{ color: filtered ? "#1677ff" : undefined }} />
```
petar-cvit, line 271: Can you set `#ff8803` here?  Its Cyclops orange 😄 

I see that we left the same blue on line 169, so could you also update that?

petar-cvit (changes requested):
Hey @m2lusen, thanks a lot for the PR. I left a couple of comments, but looks great 🧡

kubectl
m2lusen: Just removed it now

cyclops-ui/src/components/pages/NodeDetails.tsx
m2lusen: Ive changed it back to orange... honestly have no idea why i changed it to that blue color, i must of used it on a different file and copied it or something.

petar-cvit (approved):
Thanks @m2lusen, looks great 🧡

PR #392, opened 2024-07-02
fix:(#368) Removed the third unnecessary status

### LINKED ISSUE #368: Too many status indicators on deployments
### Problem

There are too many status indicators in the following image.
Untitled
### Solution

1. The first ✅ represents the **status of the module** → this one needs to stay
2. The second ✅ represents the **status of the deployment** → this one needs to stay
3. The third ✅ also represents the **status of the deployment** → **this one we can remove**

tanbirali:
closes #368 

## 📑 Description
The third status was unnecessary under deployments and hence removed.
## ✅ Checks

## ℹ Additional context
Here is an screenshot of the changes 
image


cyclops-ui/src/components/pages/ModuleDetails.tsx
```diff
@@ -510,7 +510,6 @@ const ModuleDetails = () => {
               <Title style={{ paddingRight: "10px" }} level={3}>
                 {resource.name}
               </Title>
-              {statusIcon}
```
petar-cvit, line 513: Can you also remove the code that calculates the `statusIcon`

petar-cvit (changes requested):

tanbirali:
```
      statusIcon = (
        <CheckCircleTwoTone
          style={{ verticalAlign: "middle" }}
          twoToneColor={"#52c41a"}
        />
      );
 ```
@petar-cvit  do you want me to delete this, As far as I understand the statusIcon variable is responsible for the status now, third one was unnecessary so it was removed but the other two are coming from the same variable is it necessary to remove the variable itself

petar-cvit:
There are two `statusIcon` variables, of which one is used in the header, and the other is used inside the collapse body. You need to delete the one on line 373 and remove its calculation from lines 472 to 488

petar-cvit (approved):
@tanbirali thanks, thats it 🧡

PR #395, opened 2024-07-05
 added checkbox for toggled managedFields

### LINKED ISSUE #128: Hide `managedFields` from resource manifests
When checking resources deployed by a module, for each of them, you can check the manifest of the resource fetched from the cluster. Having managed fields displayed, it is a bit harder to see the rest of the configuration, which is usually more important. The scope of the issue is to add a toggle that could hide `managedFields` from the manifest modal.

Wasiim105:
closes #128 

## ✅ Checks


cyclops-ui/src/components/pages/ModuleDetails.tsx
```diff
@@ -880,6 +908,14 @@ const ModuleDetails = () => {
         cancelButtonProps={{ style: { display: "none" } }}
         width={"40%"}
       >
+        <div style={{ marginBottom: '10px' }}>
+        <Checkbox
+        onChange={handleCheckboxChange} checked={showManagedFields} >
+          Include Managed Fields
+        </Checkbox>
+      </div>
```
petar-cvit, line 916: Could you add an [Divider](https://ant.design/components/divider) under this div?

cyclops-ui/src/components/pages/ModuleDetails.tsx
```diff
@@ -159,19 +186,20 @@ const ModuleDetails = () => {
...
           on: true,
-          manifest: res.data,
+          manifest: manifestData,
         });
       })
       .catch((error) => {
-        setLoading(false);
-        setLoadModule(true);
-        setError(mapResponseError(error));
+        console.error("Error fetching manifest:", error);
```
petar-cvit, line 200: Can you revert this? We need to set the error to the state when it occurs

cyclops-ui/src/components/pages/ModuleDetails.tsx
```diff
@@ -159,19 +186,20 @@ const ModuleDetails = () => {
...
           namespace: namespace,
+          includeManagedFields: showManagedFields,
         },
       })
       .then((res) => {
+        let manifestData = res.data;
         setManifestModal({
           on: true,
-          manifest: res.data,
+          manifest: manifestData,
```
petar-cvit, line 196: Is there a reason to add `manifestData` variable?

cyclops-ui/src/components/pages/ModuleDetails.tsx
```diff
@@ -144,6 +145,32 @@ const ModuleDetails = () => {
...
+        fetchManifest(
+          resource.group,
+          resource.version,
+          resource.kind,
+          resource.namespace,
+          resource.name,
+        );
+      }
+    })
+  },[showManagedFields]);
```
petar-cvit, line 169: I wouldn't go over all resources we have for a module until I find the one we need.

You can add `group`, `version`, `kind`, `name` and `namespace` of the resource to the `manifestModal` state when the Manifest button is clicked. This way you don't have to go over all the resources and just get it from the state.

So basically what you did with the `kind` state, but you can pack it into the same state as `manifestModal` and add all of the fields so you don't have to iterate over resources

cyclops-ctrl/internal/controller/modules.go
```diff
@@ -564,8 +564,9 @@ func (m *Modules) GetManifest(ctx *gin.Context) {
 	kind := ctx.Query("kind")
 	name := ctx.Query("name")
 	namespace := ctx.Query("namespace")
+	includeManagedFields := ctx.Query("includeManagedFields") == "false"
```
petar-cvit, line 567: Can you check if the field is set to "true"?

cyclops-ctrl/internal/cluster/k8sclient/client.go
```diff
@@ -260,6 +260,10 @@ func (k *KubernetesClient) GetManifest(group, version, kind, name, namespace str
 		return "", err
 	}
 
+	if includeManagedFields {
+		resource.SetManagedFields(nil)
+	}
+
```
petar-cvit, line 266: I left a comment about checking if the flag is equal to "true" instead of "false" in `cyclops-ctrl/internal/controller/modules.go` so could you change this to
```go
if !includeManagedFields {
  resource.SetManagedFields(nil)
}
```

petar-cvit (changes requested):

petar-cvit:
@Wasiim105 thanks for the PR, I left a couple of comments, but it works as expected

cyclops-ui/src/components/pages/ModuleDetails.tsx
```diff
@@ -862,6 +897,16 @@ const ModuleDetails = () => {
...
+        <div>
+          <Divider>
+            <Checkbox
+            onChange={handleCheckboxChange} checked={showManagedFields} >
+              Include Managed Fields
+            </Checkbox>
+          </Divider>
+        </div>
+
+
```
petar-cvit, line 909: @Wasiim105 could you add the divider below the checkbox?
```suggestion
        <div>
          <Checkbox
            onChange={handleCheckboxChange} checked={showManagedFields} >
            Include Managed Fields
          </Checkbox>
          <Divider style={{marginTop: "12px", marginBottom: "12px"}}/>
        </div>
```

cyclops-ui/src/components/pages/ModuleDetails.tsx
```diff
@@ -144,6 +152,40 @@ const ModuleDetails = () => {
 
   let { moduleName } = useParams();
 
+  const [showManagedFields, setShowManagedFields] = useState(false);
+
+
+  const handleCheckboxChange = (e: CheckboxChangeEvent) => {
+    setShowManagedFields(e.target.checked);
```
petar-cvit, line 159: You can call the fetchManifest here directly instead of using `useEffect` on line 176. Also, you have some errors because of useEffect dependencies so you can delete it and just call `fetchManifest` here

petar-cvit (changes requested):

cyclops-ui/src/components/pages/ModuleDetails.tsx
Wasiim105: @petar-cvit if I call fetchManifest  after the setShowManagedFields on line 160, it's not working as we wanted because managedFields is not updating immediately  it is working in reverse order, means if checkbox is checked then managedFields are not displayed and when we unchecked the checkbox then managedFields are displayed. I have tried from  my end but unable to solve this without useEffect, could you please guide me.

cyclops-ctrl/minikube-linux-amd64
petar-cvit, line 1: Can you also remove this one?

Wasiim105: sure

cyclops-ui/src/components/pages/ModuleDetails.tsx
petar-cvit: @Wasiim105
So I was thinking of fetching the manifest and passing it `e.target.checked` (you will also need to add that argument to the `fetchManifest` function).

```typescript
const handleCheckboxChange = (e: CheckboxChangeEvent) => {
  setShowManagedFields(e.target.checked);
  fetchManifest(
    manifestModal.resource.group,
    manifestModal.resource.version,
    manifestModal.resource.kind,
    manifestModal.resource.namespace,
    manifestModal.resource.name,
    e.target.checked,
  );
};
```

You can store it to the state (`setShowManagedFields`) so you can use it in the `handleManifestClick`. In that function you can add the fetchManifest and pass whatever was in the `showManagedFields` state

```typescript
  const handleManifestClick = (resource:any) => {
    setManifestModal({
      on: true,
      resource: {
        group: resource.group,
        version: resource.version,
        kind: resource.kind,
        name: resource.name,
        namespace: resource.namespace
      },
      manifest: ''
    });
    fetchManifest(
        resource.group,
        resource.version,
        resource.kind,
        resource.namespace,
        resource.name,
        showManagedFields,
    );
  };
```

With these changes you can delete the useEffect which requires you add a lot of dependencies which was the main reason why I am rooting for removing it. Let me know what you think

Wasiim105: Thank you for helping me. Now I know where I made a mistake.

petar-cvit (approved):
Thanks @Wasiim105 🧡

PR #412, opened 2024-07-13
disable telemetry with init

### LINKED ISSUE #364: `cyctl` telemetry flag
### Problem

When installing Cyclops with `kubectl` and the install manifests, you have the option to turn off telemetry. But when installing Cyclops with `cyctl` and the `init` command, this option does not exist. 

### Solution

Implement a `—-disable-telemetry` flag to disable telemetry when installing Cyclops via `cyctl`.

The command should look like the following:

```yaml
cyctl init —-disable-telemetry
```

thebigbone:
closes #364 

## 📑 Description
added a telemetry flag which will disable the telemetry

## ✅ Checks

### DIFF
--- a/cyctl/cmd/apply.go
+++ b/cyctl/cmd/apply.go
@@ -126,7 +126,10 @@ var applyCmd = &cobra.Command{
 			log.Fatal(err)
 		}
 
-		telemetry, _ := cmd.Flags().GetBool("disable-telemetry")
+		disableTelemetry, err := cmd.Flags().GetBool("disable-telemetry")
+		if err != nil {
+			log.Fatal(err)
+		}
 
 		var deployUrl string
 		var demoUrl string
--- a/cyctl/cmd/apply.go
+++ b/cyctl/cmd/apply.go
@@ -67,7 +67,7 @@ func doServerSideApply(ctx context.Context, cfg *rest.Config, obj *unstructured.
 	return err
 }
 
-func applyYaml(yamlFile []byte, config *rest.Config, telemetry bool) error {
+func applyYaml(yamlFile []byte, config *rest.Config, disableTelemetry bool) error {
 	multidocReader := utilyaml.NewYAMLReader(bufio.NewReader(bytes.NewReader(yamlFile)))
 
 	for {
@@ -85,7 +85,7 @@ func applyYaml(yamlFile []byte, config *rest.Config, telemetry bool) error {
 			return err
 		}
 
-		if telemetry && obj.GetKind() == "Deployment" && obj.GetName() == "cyclops-ctrl" {
+		if disableTelemetry && obj.GetKind() == "Deployment" && obj.GetName() == "cyclops-ctrl" {
 			containers, _, _ := unstructured.NestedSlice(obj.Object, "spec", "template", "spec", "containers")
 			if len(containers) > 0 {
 				container := containers[0].(map[string]interface{})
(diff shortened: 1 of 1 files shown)


cyctl/cmd/apply.go
```diff
@@ -89,6 +89,44 @@ func applyYaml(yamlFile []byte, config *rest.Config) error {
 		if err != nil {
 			return err
 		}
+
+		if telemetry {
+			if obj.GetKind() == "Deployment" {
+				labels, _, _ := unstructured.NestedMap(obj.Object, "metadata", "labels")
+				if labels != nil {
+					appLabel, _, _ := unstructured.NestedString(labels, "app")
+					if appLabel == "cyclops-ctrl" {
```
petar-cvit, line 98: Can we check for cyclops controller deployment by checking the name? Also, you can check for all of those in the same if statement
```suggestion
		if telemetry && obj.GetKind() == "Deployment" && obj.GetName() == "cyclops-ctrl" {
```

cyctl/cmd/apply.go
```diff
@@ -89,6 +89,44 @@ func applyYaml(yamlFile []byte, config *rest.Config) error {
...
+							if err != nil {
+								log.Fatal(err)
+							}
+							containers[0] = container
+							err = unstructured.SetNestedSlice(obj.Object, containers, "spec", "template", "spec", "containers")
+							if err != nil {
+								log.Fatal(err)
+							}
+						}
+						err = doServerSideApply(context.TODO(), config, obj)
```
petar-cvit, line 118: Are you first deploying the controller without telemetry and then updating it? I would prefer if just alter the object and then call the `doServerSideApply` with it

cyctl/cmd/apply.go
```diff
@@ -161,5 +201,6 @@ var applyCmd = &cobra.Command{
 
 func init() {
 	applyCmd.Flags().StringP("version", "v", "main", "specify cyclops version")
+	applyCmd.Flags().BoolP("telemetry", "t", false, "disable telemetry")
```
petar-cvit, line 204: Can we rename the flag to `disable-telemetry` and update the usage string to something like `disable emitting telemetry metrics from cyclops controller`

petar-cvit (changes requested):
Thanks @thebigbone, left a couple of comments

thebigbone:
done

cyctl/cmd/apply.go
```diff
@@ -103,6 +126,8 @@ var applyCmd = &cobra.Command{
 			log.Fatal(err)
 		}
 
+		telemetry, _ := cmd.Flags().GetBool("disable-telemetry")
```
petar-cvit, line 129: Can you handle the error and rename the variable to `disableTelemetry`?
```suggestion
		disableTelemetry, err := cmd.Flags().GetBool("disable-telemetry")
		if err != nil {
			log.Fatal(err)
		}
```

petar-cvit (changes requested):

thebigbone:
done

cyctl/cmd/apply.go
```diff
@@ -67,7 +67,7 @@ func doServerSideApply(ctx context.Context, cfg *rest.Config, obj *unstructured.
 	return err
 }
 
-func applyYaml(yamlFile []byte, config *rest.Config) error {
+func applyYaml(yamlFile []byte, config *rest.Config, telemetry bool) error {
```
petar-cvit, line 70: Can you also rename this one?

petar-cvit (changes requested):

cyctl/cmd/apply.go
thebigbone: done

petar-cvit (approved):
Thanks @thebigbone 🧡

PR #414, opened 2024-07-13
feat: add status description for pods

### LINKED ISSUE #197: Add status description for pods
If a pod is not started successfully, there should be a more detailed description in the UI about why it didn't start

SpiffyEight77:
closes #197

## ✅ Checks

### Screenshot

Screenshot 2024-07-14 at 15 23 10


petar-cvit (changes requested):
Hey @SpiffyEight77, thanks for the PR, the tooltip you added looks great.
Would it make sense to add the tooltip to the `image` column so users can check the statues of containers independently? This way we also don't need an additional column

SpiffyEight77:
@petar-cvit Thank you for reviewing this PR. I will remove the `status` column and add the tooltip to the `image` column later tonight. 🙇🏻‍♂️

SpiffyEight77:
Add a tooltip with the pod description to the image column, as shown in the following screenshot.

image

petar-cvit (approved):
Thanks @SpiffyEight77 🧡

PR #437, opened 2024-07-16
Health filter for modules

nikhiltanna:
closes #

## ✅ Checks


petar-cvit:
Hey @nikhiltanna, thanks for the PR.

Can we make the filter look like the image below? You can add the statuses as checkboxes, which allows filtering multiple statuses at once.

Screenshot 2024-07-17 at 11 15 40

You can check how we have implemented it for filtering resources [here](https://github.com/cyclops-ui/cyclops/blob/773319218ac80ba41706e7800356ca11800263c8/cyclops-ui/src/components/pages/ModuleDetails.tsx#L859).

nikhiltanna:
Hey @petar-cvit  can you please check this now I have changed to popover if you can approve and merge it would be great

nikhiltanna:
Hey @petar-cvit  have changed it to popover as asked can you please approve it.

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -26,10 +28,23 @@ const Modules = () => {
   const [allData, setAllData] = useState([]);
   const [filteredData, setFilteredData] = useState([]);
   const [loadingModules, setLoadingModules] = useState(false);
+  const [displaymodule, setDisplayModule] = useState<string[]>([]);
+  const [resourceFilter, setResourceFilter] = useState<string[]>([
+    "ALL Modules",
+    "Healty",
+    "Unhealthy",
+    "Unknown",
```
petar-cvit, line 36: Module status is lowercase so this wouldn't match

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -88,108 +125,207 @@ const Modules = () => {
...
+                    )}
+                  </span>
+                </Col>
+              </Row>
+            </Card>
+          </a>
+        </Col>
+      ));
+    }
+    return filteredData
```
petar-cvit, line 230: Can we do the filtering when checkboxes change and save the filtered modules into the `filteredData`? That way, you don't have to change the part that represents the modules.

Check `handleSearch` function from the same file

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -227,14 +363,20 @@ const Modules = () => {
...
-            style={{ width: "30%" }}
+            style={{ width: "100%" }}
             onChange={handleSearch}
           ></Input>
         </Col>
+        <Col span={2}>
+          <Popover content={resourceFilterPopover()} trigger={["click"]}>
+            <Button>{"Module Type "}</Button>
+          </Popover>
+        </Col>
```
petar-cvit, line 379: Can you make this button just an extension of the search bar like in this [example](https://codepen.io/pen?&editors=001)

petar-cvit (changes requested):
You should merge main into your branch. The build is failing because of some dependencies.

Can we remove the "all modules" option? All modules should be shown if none of the options are selected

cyclops-ui/src/components/pages/Modules/Modules.tsx
nikhiltanna: This is handled during checking or returning the the value

petar-cvit: This should trigger when checkboxes change

cyclops-ui/src/components/pages/Modules/Modules.tsx
nikhiltanna: I think link is not working

petar-cvit: Oh yeah sorry, I can't share codepen links. You can check out [antd input](https://ant.design/components/input) and how to add the search icon to the right

Screenshot 2024-07-22 at 22 26 22

nikhiltanna:
Hey @petar-cvit have changed it and made it look like search extension

cyclops-ui/src/components/pages/Modules/Modules.tsx
nikhiltanna: Changed it and let me know if CSS changes are required

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -26,6 +28,9 @@ const Modules = () => {
   const [allData, setAllData] = useState([]);
   const [filteredData, setFilteredData] = useState([]);
   const [loadingModules, setLoadingModules] = useState(false);
+  const [displaymodule, setDisplayModule] = useState<string[]>([]);
+  const resourceFilter = ["Healty", "Unhealthy", "Unknown"];
+  const [isHovered, setIsHovered] = useState(false);
```
petar-cvit, line 33: You don't need this state. If you need to update the component style if the component is hovered, you can use CSS

petar-cvit (changes requested):

cyclops-ui/src/components/pages/Modules/Modules.tsx
nikhiltanna: Removed state

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -241,14 +363,35 @@ const Modules = () => {
...
+              width: "100%",
+              borderTopRightRadius: "0px",
+              borderBottomRightRadius: "0px",
+            }}
             onChange={handleSearch}
           ></Input>
         </Col>
+        <Col span={2}>
+          <Popover content={resourceFilterPopover()} trigger={["click"]}>
+            <Button
```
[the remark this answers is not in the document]
nikhiltanna: Made the required changes

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -107,103 +131,200 @@ const Modules = () => {
...
-              >
-                Path:
-                <Link
-                  aria-level={3}
-                  target="_blank"
-                  href={
-                    module.template.repo +
-                    `/tree/` +
-                    getLinkPath(
+    if (displaymodule.length === 0) {
```
petar-cvit, line 134: Check this comment https://github.com/cyclops-ui/cyclops/pull/437#discussion_r1686814924

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -241,14 +362,27 @@ const Modules = () => {
...
+            addonAfter={
+              <>
+                <Popover content={resourceFilterPopover()} trigger="click">
+                  <FilterOutlined />
+                </Popover>
+              </>
+            }
           ></Input>
         </Col>
+        <Col span={2}></Col>
```
petar-cvit, line 385: You can remove this

petar-cvit (changes requested):

cyclops-ui/src/components/pages/Modules/Modules.tsx
nikhiltanna: Have changed the code accordingly

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -45,11 +49,41 @@ const Modules = () => {
...
+      <Checkbox.Group
+        key={index}
+        style={{ display: "block" }}
+        onChange={(checkedValues) => handleSelectItem(item)}
+        value={displaymodule}
+      >
+        <>
+          <Checkbox value={item}>{item}</Checkbox>
+        </>
+      </Checkbox.Group>
```
petar-cvit, line 84: Does this create a new checkbox group for each resourceFilter? If yes, we should have a single group with multiple checkboxes

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -164,11 +193,10 @@ const Modules = () => {
                 Path:
                 <Link
                   aria-level={3}
-                  target="_blank"
                   href={
                     module.template.repo +
                     `/tree/` +
-                    getLinkPath(
+                    getTemplateVersion(
```
petar-cvit, line 199: Can you revert this part?

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -26,6 +28,8 @@ const Modules = () => {
   const [allData, setAllData] = useState([]);
   const [filteredData, setFilteredData] = useState([]);
   const [loadingModules, setLoadingModules] = useState(false);
+  const [displaymodule, setDisplayModule] = useState<string[]>([]);
```
petar-cvit, line 31: Can you rename this to `moduleHealthFilter`?

petar-cvit (changes requested):
Can you revert the deleted empty lines? Also, the build in CI fails because the  useEffect needs the filteredData as the dependency (alongside displaymodule)

cyclops-ui/src/components/pages/Modules/Modules.tsx
nikhiltanna: Done

cyclops-ui/src/components/pages/Modules/Modules.tsx
nikhiltanna: changed according to the requirements

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -45,11 +49,40 @@ const Modules = () => {
...
+  useEffect(() => {
+    setFilteredData(
+      filteredData.filter((module: any) =>
+        moduleHealthFilter
+          .map((status) => status.toLowerCase())
+          .includes(module.status.toLowerCase()),
+      ),
+    );
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [moduleHealthFilter]);
```
petar-cvit, line 61: @nikhiltanna, lets add the missing dependency instead of silencing it

petar-cvit (changes requested):

cyclops-ui/src/components/pages/Modules/Modules.tsx
nikhiltanna: If we add it will keep on retriggering and will set the infinite loop of renders

nikhiltanna: Have updated the dependency array as well sorry for the issue

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -143,10 +174,7 @@ const Modules = () => {
                 }}
               >
                 Repo:
-                <Link aria-level={3}
-                 href={module.template.repo}
-                 target="_blank"
-                 >
+                <Link aria-level={3} href={module.template.repo}>
```
petar-cvit, line 177: Can you revert this part?

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -164,7 +192,6 @@ const Modules = () => {
                 Path:
                 <Link
                   aria-level={3}
-                  target="_blank"
```
petar-cvit, line 167: Can you revert this part?

nikhiltanna:
Are there any other changes required?

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -45,11 +49,40 @@ const Modules = () => {
...
+    const newfilteredData = filteredData.filter((module: any) =>
+      moduleHealthFilter
+        .map((status) => status.toLowerCase())
+        .includes(module.status.toLowerCase()),
+    );
+
+    if (JSON.stringify(newfilteredData) !== JSON.stringify(filteredData)) {
+      setFilteredData(newfilteredData);
+    }
+  }, [moduleHealthFilter, filteredData]);
```
petar-cvit, line 62: I tried it locally, and it doesn't seem to work. The problem is that when you unselect one health filter and select it again, it will not show the modules with that health status. This happens because you are always filtering on top of existing `filteredData`, which doesn't contain all modules but only the ones that passed the previous filter.

I would propose a different solution. You can have a similar `useEffect` that has `moduleHealthFilter` as a dependency and the input to the search field as a dependency. This will require you to store the search input text as a separate state (you can call this other state `searchInputFilter`), and the `handleSearch` function can simply store the search input in the state.

When you have both `moduleHealthFilter` and `searchInputFilter` (search input text), you can filter based on the health as you did above and then apply the filter like it's implemented in the `handleSearch` now. After filtering, you can simply store it in the `filteredData`.

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -26,6 +28,8 @@ const Modules = () => {
   const [allData, setAllData] = useState([]);
   const [filteredData, setFilteredData] = useState([]);
   const [loadingModules, setLoadingModules] = useState(false);
+  const [moduleHealthFilter, setModuleHealthFilter] = useState<string[]>([]);
+  const resourceFilter = ["Healty", "Unhealthy", "Unknown"];
```
petar-cvit, line 32:
```suggestion
  const resourceFilter = ["Healthy", "Unhealthy", "Unknown"];
```

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -26,6 +28,8 @@ const Modules = () => {
   const [allData, setAllData] = useState([]);
   const [filteredData, setFilteredData] = useState([]);
   const [loadingModules, setLoadingModules] = useState(false);
+  const [moduleHealthFilter, setModuleHealthFilter] = useState<string[]>([]);
```
petar-cvit, line 31: You should start with all health filters enabled

```suggestion
  const [moduleHealthFilter, setModuleHealthFilter] = useState<string[]>([
    "healthy",
    "unhealthy",
    "unknown",
  ]);
```

petar-cvit (changes requested):
@nikhiltanna I proposed a different approch since this one did not work. Check comments below

cyclops-ui/src/components/pages/Modules/Modules.tsx
nikhiltanna: Done

cyclops-ui/src/components/pages/Modules/Modules.tsx
nikhiltanna: I have tried something can you please check once now it would be great

cyclops-ui/src/components/pages/Modules/Modules.tsx
petar-cvit: @nikhiltanna can you also revert this one

cyclops-ui/src/components/pages/Modules/Modules.tsx
petar-cvit: @nikhiltanna, you added `allData` as a dependency, but this way, we are performing filters in separate functions. In the solution suggested above, filters are applied in the same place, so it's easier to read and maintain.

petar-cvit (changes requested):
@nikhiltanna I left another comment [here](https://github.com/cyclops-ui/cyclops/pull/437#discussion_r1714341049)

cyclops-ui/src/components/pages/Modules/Modules.tsx
nikhiltanna: I made the changes as asked and thanks for the suggestion it really makes code readable as all filter functions gather into one place

cyclops-ui/src/components/pages/Modules/Modules.tsx
nikhiltanna: Done

cyclops-ui/src/components/pages/Modules/Modules.tsx
```diff
@@ -45,18 +54,50 @@ const Modules = () => {
...
+    const newfilteredData = updatedList.filter((module: any) =>
+      moduleHealthFilter
+        .map((status) => status.toLowerCase())
+        .includes(module.status.toLowerCase()),
+    );
+
+    if (JSON.stringify(newfilteredData) !== JSON.stringify(filteredData)) {
+      setFilteredData(newfilteredData);
+    }
+  }, [moduleHealthFilter, filteredData, allData, searchInputFilter]);
```
petar-cvit, line 74: @nikhiltanna, can you remove `filteredData` from the dependencies?

petar-cvit (changes requested):
@nikhiltanna I left one last comment. Should be good to merge after that

cyclops-ui/src/components/pages/Modules/Modules.tsx
nikhiltanna: Done please check if it is working fine now

nikhiltanna: Should I add the dependency or silent it??

nikhiltanna: Even if we keep filerdata as dependency it's will not  affect as I have kept if conditions to update the filterdata  both ways  it would work actually

petar-cvit: @nikhiltanna Actually, if you remove the condition (the one with JSON.stringify) and just set filtered data you should be able to remove filteredData from dependencies. I prefer we don't have filteredData there because each time you change it, it will trigger the useEffect for nothing

nikhiltanna: Done

petar-cvit (approved):
Thanks @nikhiltanna 🧡

PR #449, opened 2024-07-19
⛏ fix(cyclops-ui): double err alerts on template addition and update!

### LINKED ISSUE #284: Duplicate Error Warning on Templates Screen
## Problem

When you try to add a new template to Cyclops and fail, the error is displayed both on the modal and in the background (as the picture below shows). This is unnecessary, and we should only display the error on the modal.

Screenshot 2024-05-21 at 14 24 24

### Solution

Implement a separate error state for the modal in `cyclops-ui/src/components/TemplateStore/TemplateStore.tsx`

The `Alert` on line 278 represents the background error and the `Alert` on line 359 represents the error shown on the modal.

### Alternative Solution

Implement error notifications similar to this [PR](https://github.com/cyclops-ui/cyclops/pull/280)

ArnavK-09:
closes #284

## 📑 Description
image

## ✅ Checks

## ℹ Additional context

so currently implemented antd notifications error for only template addition and update.
am currently confused should I add for every actions for template? like delete validate etc.
also is this way correct?

and should i remove `setError(mapResponseError(error));` and `<Alert/>` in modal?

### DIFF
--- a/cyclops-ui/src/components/pages/TemplateStore/TemplateStore.tsx
+++ b/cyclops-ui/src/components/pages/TemplateStore/TemplateStore.tsx
@@ -76,7 +76,7 @@ const TemplateStore = () => {
     errors.forEach(function (error: any) {
       errorMessages.push({
         key: error.message,
-        errors: error.description,
+        errors: [error.description],
       });
     });
     openNotification(errorMessages);
--- a/cyclops-ui/src/components/pages/TemplateStore/TemplateStore.tsx
+++ b/cyclops-ui/src/components/pages/TemplateStore/TemplateStore.tsx
@@ -11,6 +11,7 @@ import {
   Divider,
   message,
   Spin,
+  notification,
 } from "antd";
 import axios from "axios";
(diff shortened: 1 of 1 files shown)


KaradzaJuraj (changes requested):
Yes, @ArnavK-09, this is what we wanted! However, the error message description is not readable because it lacks spaces between words. Could you look into that?

> should I add for every actions for template? like delete validate etc

Implement it for `validate` as well.


> and should i remove setError(mapResponseError(error)); and <Alert/> in modal?

Don't remove that `<Alert>` as it is used when the backend is not working.

ArnavK-09:
> However, the error message description is not readable because it lacks spaces between words. Could you look into that?

image

um, now i fixed spacing and now its correctly showing api err desc with spacing

anything i have to do?

ArnavK-09:
#### before

image


#### after

image

KaradzaJuraj (approved):
Awesome @ArnavK-09 🙌

Issue #439, opened 2024-07-17
Add `template` flag to cyctl on module creation


petar-cvit:
**Is your feature request related to a problem? Please describe.**
When creating a module using cyctl, you need to provide flags to define the repo, path, and version of the template.

The scope of the issue is to add a flag called template, which accepts a string. If provided, `cyctl` needs to fetch the template store custom resource with the name provided via the `template` flag. Once the template store is fetched, repo, path and version defined in the template store are used for the module creation the same way its done with the flags.

You can find how to fetch a template store custom resource [here](https://github.com/cyclops-ui/cyclops/blob/ceb523172ef0a8e9845ea19f34e63d1559f5e220/cyctl/internal/describe/template_store.go#L24).

The solution should not remove the repo, path and version flags, but just add a new one. Additionally, template flag cant be combined none of the flags for repo, path and version.

shivamrazorpay:
@petar-cvit if we provide template flag & (repo, path & version) flag. Will the values from template over ride the values provided in the flag or we should have a validation that either template flag is there or the other (repo,path & version) flag ??

petar-cvit:
Hey @shivamrazorpay, lets throw an error if the template flag is set along repo, path or version

shivamrazorpay:
@petar-cvit I have created a PR, but I am new to the repo, so not sure how to locally run the CLI.
Can you give me some resources? or you can review my Code.
Thank You.

petar-cvit:
You should set up a k8s cluster and then you should be able to create modules there. To create a cluster on your local machine, you can follow out [contributing docs](https://github.com/cyclops-ui/cyclops/blob/main/CONTRIBUTING.md#cluster-setup)

petar-cvit:
implemented in #456

PR #456, opened 2024-07-20
Feat :Add template flag to cyctl on module creation

shivamrazorpay:
closes #

## 📑 Description
- Added Template Flag to cycle on Module Creation.
- Fixes Issue : #439 

## ✅ Checks

### DIFF
--- a/cyctl/internal/create/modules.go
+++ b/cyctl/internal/create/modules.go
@@ -22,11 +22,12 @@ var (
 	--path '/path/to/charts' \
 	--version 'main'
 	`
-	valuesFile string
+	valuesFile   string
+	templateName string
 )
 
 // createModule allows you to create module Custom Resource.
-func createModule(clientset *client.CyclopsV1Alpha1Client, moduleName, repo, path, version, namespace, valuesFile string) {
+func createModule(clientset *client.CyclopsV1Alpha1Client, moduleName, repo, path, version, namespace, valuesFile, tempName string) {
 
 	values, err := os.ReadFile(valuesFile)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cyctl/internal/create/modules.go
```diff
@@ -74,7 +86,12 @@ var (
 		Args:    cobra.ExactArgs(1),
 		Aliases: []string{"modules"},
 		Run: func(cmd *cobra.Command, args []string) {
-			createModule(kubeconfig.Moduleset, args[0], repo, path, version, namespace, valuesFile)
+			// Custom validation:
+			// Either templateName or (repo and path) must be provided, if one is provided the other must be empty
+			if (templateName != "" && (repo != "" || path != "")) || (templateName == "" && (repo == "" || path == "")) {
+				log.Fatalf("Error: Either template or (repo and path) must be provided.")
+			}
```
petar-cvit, line 93: Can you also add version here?
```suggestion
			if (templateName != "" && (repo != "" || path != "" || version != "")) || (templateName == "" && (repo == "" || path == "")) {
				log.Fatalf("Error: Either template or (repo, path and version) must be provided.")
			}
```

petar-cvit (changes requested):
Left one comment. Thanks @shivamrazorpay for the PR

cyctl/internal/create/modules.go
shivamrazorpay: Version was not a compulsory flag that is why I left it. I have added it now.

petar-cvit: Makes sense @shivamrazorpay. This is just users dont use the --version flag to override the version in the template.
Can you push the update?

petar-cvit (approved):
Thanks @shivamrazorpay 🧡

PR #460, opened 2024-07-21
Add new style to modules button

Wasiim105:
closes #

## ✅ Checks


petar-cvit (changes requested):
Hey @Wasiim105, can you actually revert the ModuleDetails page? Now that I look at it, the orange and red really pop up. We don't want people to panic when they see a lot of orange and red on the module details page where you'd usually want everything to be green.

petar-cvit (approved):
Thanks @Wasiim105 🧡

PR #461, opened 2024-07-21
use the latest version for the OCI template default version

pithecuse527:
closes #

## 📑 Description

If the targetVersion is empty, the resolveSemver sets it to the constraint "> 0.0.0".

This ensures that any non-empty version will match, allowing the function to return the highest available version.

## ✅ Checks


petar-cvit (approved):
Thanks @pithecuse527 🧡

PR #474, opened 2024-07-28
🚀 feat: added cache metrics to the prometheus monitor

### LINKED ISSUE #469: Expose template cache metrics to prometheus
Cyclops controller fetches templates (helm charts) to render UIs and deploy applications. To improve the performance of template fetching, the controller uses an in-memory cache for templates.

The goal of the issue is to expose metrics from the template cache to the prometheus server and have them exposed on `/metrics`.

Cache is defined [here](https://github.com/cyclops-ui/cyclops/blob/main/cyclops-ctrl/internal/template/cache/inmemory.go) as `ristretto.Cache` and metrics are available in the `Metrics` field. [Here](https://github.com/cyclops-ui/cyclops/blob/main/cyclops-ctrl/internal/prometheus/handler.go) is the Prometheus monitor where we should add new metrics for the templates cache and expose methods for updating metric values.

Cache metrics can be exposed each time a new cache request is initiated or periodically using a cronjob that runs in a separate goroutine.

naineel1209:
closes #469

## 📑 Description
- added cache metrics to the Prometheus monitor
- currently added metrics are 
```go
 type Monitor struct {
	ModulesDeployed  prometheus.Gauge
	CacheHits        prometheus.Counter
	CacheMisses      prometheus.Counter
	CacheKeysAdded   prometheus.Counter
	CacheCostAdded   prometheus.Counter
	CacheKeysEvicted prometheus.Counter
	CacheCostEvicted prometheus.Counter
}
```
- added methods to update the cache metrics
- added cronjob which run every 10 seconds to update the cache metrics

## ✅ Checks

## ℹ Additional context
- More metrics for the cache can be added if and when required
- Current time delta of 10 seconds to update the cache metrics is chosen randomly

### DIFF
--- a/cyclops-ctrl/internal/prometheus/handler.go
+++ b/cyclops-ctrl/internal/prometheus/handler.go
@@ -1,26 +1,80 @@
 package prometheus
 
 import (
+	"time"
+
+	"github.com/dgraph-io/ristretto"
 	"github.com/go-logr/logr"
 	"github.com/prometheus/client_golang/prometheus"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
 type Monitor struct {
-	ModulesDeployed prometheus.Gauge
+	ModulesDeployed  prometheus.Gauge
+	CacheHits        prometheus.Counter
+	CacheMisses      prometheus.Counter
+	CacheKeysAdded   prometheus.Counter
+	CacheCostAdded   prometheus.Counter
+	CacheKeysEvicted prometheus.Counter
+	CacheCostEvicted prometheus.Counter
 }
 
 func NewMonitor(logger logr.Logger) (Monitor, error) {
+
 	m := Monitor{
 		ModulesDeployed: prometheus.NewGauge(prometheus.GaugeOpts{
 			Name:      "modules_deployed",
 			Help:      "No of modules Inc or Dec",
 			Namespace: "cyclops",
 		}),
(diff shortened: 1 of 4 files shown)


cyclops-ctrl/internal/prometheus/handler.go
```diff
@@ -33,3 +87,27 @@ func (m *Monitor) IncModule() {
 func (m *Monitor) DecModule() {
 	m.ModulesDeployed.Dec()
 }
+
+func (m *Monitor) UpdateCacheMetrics(cache *ristretto.Cache) {
+	cacheMetrics := cache.Metrics
+
+	m.CacheHits.Add(float64(cacheMetrics.Hits()))
```
petar-cvit, line 94: Shouldn't we set cacheMetrics values instead of adding them since cacheMetrics will give you the exact value of the metric, not just the diff? This also requires to update metrics from counters to gauges

petar-cvit (changes requested):
Thanks @naineel1209 for the PR. I left one comment on metrics type

cyclops-ctrl/internal/prometheus/handler.go
naineel1209: resolved in: 43b5a46f6d45079f5ec86c80487a33b6e5c5893d

- changed the cache metric type in monitor to `prometheus.Gauge`
- refactored to set the latest cache metric instead of adding to the cache metrics

petar-cvit (approved):
Thanks @naineel1209 🧡

PR #478, opened 2024-07-30
feat: Add loading state for template in EditModule and NewModule components

### LINKED ISSUE #423: Disable all actions while the template is loading
When trying to add a new module, you have to choose one of your templates. Depending on the template size, it could take some time to load the template.

We should disable all user actions while the template is loading in:

- disable the `save` button
- disable the `load values from file` button
- disable changing the template

ChiragAgg5k:
closes #423 

## 📑 Description

Disables the `save` and `load values from file` buttons when a template is loading. also, it Disables changing of template when a template is loading.

## ✅ Checks

### DIFF
--- a/cyclops-ui/src/components/pages/EditModule/EditModule.tsx
+++ b/cyclops-ui/src/components/pages/EditModule/EditModule.tsx
@@ -90,6 +90,7 @@ const EditModule = () => {
   const [values, setValues] = useState({});
   const [isChanged, setIsChanged] = useState(false);
   const [isTemplateChanged, setIsTemplateChanged] = useState(false);
+  const [isTemplateLoading, setIsTemplateLoading] = useState(false);
   const [config, setConfig] = useState<Template>({
     name: "",
     resolvedVersion: "",
@@ -308,6 +309,7 @@ const EditModule = () => {
   };
 
   async function handleSubmitTemplateEdit(templateEditValues: any) {
+    setIsTemplateLoading(true);
     setLoadTemplate(false);
 
     let currentValues = form.getFieldsValue();
@@ -378,6 +380,7 @@ const EditModule = () => {
     setInitialValuesRaw(initialValuesResult.initialValues);
     form.setFieldsValue(mergedValuesMapped);
 
+    setIsTemplateLoading(false);
     setLoadTemplate(true);
   }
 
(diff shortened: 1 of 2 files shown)


cyclops-ui/src/components/pages/EditModule/EditModule.tsx
petar-cvit, line 1: Can we use existing `loadTemplate` instead of adding a new variable `isTemplateLoading`? `loadTemplate` indicates if a template has been loaded, so if it's false, it means the template is still loading.

petar-cvit (changes requested):
@ChiragAgg5k thanks for the PR. Left one comment on reusing existing state we have in edit module

cyclops-ui/src/components/pages/EditModule/EditModule.tsx
ChiragAgg5k: Oh, sorry for the oversight. I will make the changes accordingly

cyclops-ui/src/components/pages/NewModule/NewModule.tsx
```diff
@@ -1283,18 +1284,24 @@ const NewModule = () => {
...
               >
                 Load values from file
               </Button>{" "}
               <Button
                 type="primary"
-                loading={loading}
+                loading={loading || loadingTemplate}
                 htmlType="submit"
                 name="Save"
+                disabled={loadingTemplate}
```
petar-cvit, line 1296: Can you also disable the button if `loadingTemplateInitialValues` is set? This is set to true while Cyclops is fetching initial values from a template. You can do it for all buttons

cyclops-ui/src/components/pages/EditModule/EditModule.tsx
```diff
@@ -1227,7 +1228,10 @@ const EditModule = () => {
                 name={"repo"}
                 style={{ width: "40%", marginRight: "0" }}
               >
-                <Input placeholder={"Repository"} disabled={templateRefLock} />
+                <Input
+                  placeholder={"Repository"}
+                  disabled={templateRefLock || !loadTemplate}
```
petar-cvit, line 1233: You should add this to the path input as well.

petar-cvit (changes requested):

cyclops-ui/src/components/pages/NewModule/NewModule.tsx
petar-cvit: @ChiragAgg5k, you will have to add loadingTemplateInitialValues to disabled as well.

petar-cvit (changes requested):

petar-cvit (approved):
Thanks @ChiragAgg5k 🧡

PR #480, opened 2024-07-30
Implements persistentVolume backend

### LINKED ISSUE #434: PersistentVolume Backend Implementation

```bash
```

The goal of the issue is to support PersistentVolumes on the same endpoint. The endpoint for fetching PersistentVolume data would look like this: 

```bash
 /resources?group=apps&version=v1&kind=PersistentVolume&name=&namespace=
```

The endpoint should return data about its `name`, `capacity`, `accessModes`, `status`, `persistentVolumeClaim`, `reclaimPolicy`, `strageClass`… (you can find more about it [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#persistentvolume-v1-core))

jplourenco1:
closes #434

## ✅ Checks

### DIFF
--- a/cyclops-ctrl/internal/cluster/k8sclient/client.go
+++ b/cyclops-ctrl/internal/cluster/k8sclient/client.go
@@ -347,6 +347,8 @@ func (k *KubernetesClient) GetResource(group, version, kind, name, namespace str
 		return k.mapPod(group, version, kind, name, namespace)
 	case isConfigMap(group, version, kind):
 		return k.mapConfigMap(group, version, kind, name, namespace)
+	case isPersistentVolume(group, version, kind):
+		return k.mapPersistentVolumes(group, version, kind, name, namespace)
 	case isPersistentVolumeClaims(group, version, kind):
 		return k.mapPersistentVolumeClaims(group, version, kind, name, namespace)
 	case isSecret(group, version, kind):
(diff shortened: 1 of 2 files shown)


cyclops-ctrl/internal/cluster/k8sclient/client.go
```diff
@@ -772,6 +774,32 @@ func (k *KubernetesClient) mapPersistentVolumeClaims(group, version, kind, name,
...
+	}
+
+	return &dto.PersistentVolume{
+		Group:                 group,
+		Version:               version,
+		Kind:                  kind,
+		Name:                  name,
+		Namespace:             namespace,
+		AccessModes:           persistentvolume.Spec.AccessModes,
+		PersistentVolumeClaim: persistentvolume.Spec.ClaimRef.Name,
```
petar-cvit, line 795: This can also produce a nil pointer dereference since `ClaimRef` is a pointer. Can you solve it the same way as capacity?

cyclops-ctrl/internal/cluster/k8sclient/client.go
```diff
@@ -772,6 +774,32 @@ func (k *KubernetesClient) mapPersistentVolumeClaims(group, version, kind, name,
 	}, nil
 }
 
+func (k *KubernetesClient) mapPersistentVolumes(group, version, kind, name, namespace string) (*dto.PersistentVolume, error) {
+	persistentvolume, err := k.clientset.CoreV1().PersistentVolumes().Get(context.Background(), name, metav1.GetOptions{})
```
petar-cvit, line 778: Can you rename this one to `persistentVolume`?

petar-cvit (changes requested):
Thanks for the PR @jplourenco1. Left two comments

cyclops-ctrl/internal/cluster/k8sclient/client.go
jplourenco1: fixed using "Retain" as a default policy

cyclops-ctrl/internal/cluster/k8sclient/client.go
jplourenco1: fixed as suggested

cyclops-ctrl/internal/cluster/k8sclient/client.go
jplourenco1: fixed as suggested

petar-cvit (approved):
Thanks @jplourenco1 🧡

PR #491, opened 2024-08-02
changed state management for loading

### LINKED ISSUE #421: Validating Multiple Templates Rotates Only The Last Icon
In the templates tab, it is possible to start the validation process for multiple templates simultaneously, but the icon will change only for the last one you started.

We want the icon to be changed for every template that is currently being validated.

nipunh:
closes #421 

## ✅ Checks
- [*] I have updated the documentation as required
- [* ] I have performed a self-review of my code


petar-cvit (approved):
Thanks @nipunh 🧡

PR #493, opened 2024-08-03
implement networkpolicy data api for controller

### LINKED ISSUE #433: NetworkPolicy Backend Implementation

If you want to fetch information about a Deployment from default namespace called demo-app you would call:

```bash
```

The goal of the issue is to support NetworkPolicies on the same endpoint. The endpoint for fetching NetworkPolicies data would look like this: 

```bash
 /resources?group=apps&version=v1&kind=NetworkPolicy&name=&namespace=
```

The endpoint should return data about its rules, you can find more about it [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#networkpolicy-v1-networking-k8s-io)

A structure similar to this one is preferred: 

```bash
 {”ingress”: [array_of_rules], ”egress”:[array_of_rules]}
```

pithecuse527:
closes #433 

## 📑 Description
- This PR implements fetching data for NetworkPolicy on the controller.
- Here is an example for the API call:
  - `/resources?group=networking.k8s.io&version=v1&kind=NetworkPolicy&name=test-network-policy&namespace=default`
- Here is an example for the API returns:
	```json
	{
	  "group": "networking",
	  "version": "v1",
	  "kind": "NetworkPolicy",
	  "name": "test-network-policy",
	  "namespace": "default",
	  "pods": [],
	  "ingress": [
	    {
	      "ports": [
	        {
	          "protocol": "TCP",
	          "port": 6379,
	          "endPort": 0
	        }
	      ],
	      "from": [
	        {
	          "ipBlock": {
	            "cidr": "172.17.0.0/16",
	            "except": [
	              "172.17.1.0/24"
	            ]
	          }
	        },
	        {
	          "ipBlock": null
	        },
	        {
	          "ipBlock": null
	        }
	      ]
	    }
	  ],
	  "egress": [
	    {
	      "ports": [
	        {
	          "protocol": "TCP",
	          "port": 5978,
	          "endPort": 0
	        }
	      ],
	      "to": [
	        {
	          "ipBlock": {
	            "cidr": "10.0.0.0/24",
	            "except": null
	          }
	        }
	      ]
	    }
	  ]
	}
	
	```

## ✅ Checks

### DIFF
--- a/cyclops-ctrl/internal/cluster/k8sclient/client.go
+++ b/cyclops-ctrl/internal/cluster/k8sclient/client.go
@@ -7,6 +7,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	networkingv1 "k8s.io/api/networking/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"os"
 	"os/exec"
 	"sort"
@@ -359,6 +361,8 @@ func (k *KubernetesClient) GetResource(group, version, kind, name, namespace str
 		return k.mapJob(group, version, kind, name, namespace)
 	case isRole(group, version, kind):
 		return k.mapRole(group, version, kind, name, namespace)
+	case isNetworkPolicy(group, version, kind):
+		return k.mapNetworkPolicy(group, version, kind, name, namespace)
 	}
 
 	return nil, nil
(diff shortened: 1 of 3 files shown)


cyclops-ctrl/internal/cluster/k8sclient/client.go
```diff
@@ -953,3 +1050,7 @@ func isSecret(group, version, kind string) bool {
 func isCronJob(group, version, kind string) bool {
 	return group == "batch" && version == "v1" && kind == "CronJob"
 }
+
+func isNetworkPolicy(group, version, kind string) bool {
+	return group == "networking" && version == "v1" && kind == "NetworkPolicy"
```
petar-cvit, line 1055: Should the group be `networking.k8s.io`?

petar-cvit (changes requested):
Thanks @pithecuse527, left one comment. Great job on this, you really added all there is to add on network policies. Love that we also have pods mapped 🧡

cyclops-ctrl/internal/cluster/k8sclient/client.go
pithecuse527: Yes, it should.
Need to check again.

pithecuse527: Changed the group name with "networking.k8s.io"

cyclops-ctrl/internal/models/dto/k8s.go
```diff
@@ -17,6 +18,13 @@ type Resource interface {
 	SetDeleted(bool)
 }
 
+type PolicyType string
+
+const (
+	PolicyTypeIngress PolicyType = "Ingress"
+	PolicyTypeEgress  PolicyType = "Egress"
+)
+
```
petar-cvit, line 27: Are these used?

petar-cvit (changes requested):
@pithecuse527 left one more comment

cyclops-ctrl/internal/models/dto/k8s.go
pithecuse527: unused constants removed

petar-cvit (approved):
Thanks @pithecuse527 🧡

PR #496, opened 2024-08-06
scope cyclops controller to single namespace

### LINKED ISSUE #345: Scope cyclops controller to a namespace
Kubernetes controllers listen to changes on certain resource types. In the case of Cyclops, the controller is listening to changes on the Module resource. Currently, modules can be deployed to any namespace, but the controller should only take those deployed to the `cyclops` namespace.

The goal of the issue is to scope the controller just to a single namespace. By default it should be `cyclops`, but a user should be able to configure it using an environment variable.

You can check how to scope a controller to a single namespace [here](https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/). Also, the definition of our controller options is [here](https://github.com/cyclops-ui/cyclops/blob/ba0a74443cfb82ffc05bf9ba58b22a4d27fbb1ee/cyclops-ctrl/cmd/main/main.go#L90).

kevalsabhani:
closes #345

## 📑 Description
Added a property `Cache` while creating `manager` to scope cyclops controller to single namespace. Namespace will be taken as an env variable, but if it is empty, it will be defaulted to `cyclops`.

## ✅ Checks


petar-cvit (changes requested):
Thanks for the PR @kevalsabhani!

When a module reconciliation is triggered, the Reconciler is [fetching the module](https://github.com/cyclops-ui/cyclops/blob/90f2d765e14babd942940909efde73813033ab90/cyclops-ctrl/internal/modulecontroller/module_controller.go#L198) once more, but always from the cyclops namespace, which fails if the module is deployed in a different namespace. Could you also fix that part so the module fetched in the beginning of the reconciliation func is passed to `moduleToResources` func instead of fetching the module again?

Apart from that, seems like reconciler is picking up changes from the namespaces provided in the env var - good job!

kevalsabhani:
Hi @petar-cvit , I have the function as per the suggestion, please review. Thanks.

petar-cvit (approved):
Thanks @kevalsabhani 🧡

PR #505, opened 2024-08-12
Disabling save button if no template is selected

### LINKED ISSUE #503: Disable save button before selecting a template
When creating a new Module, you can choose templates from the dropdown. However, even before selecting the template, the Save button is active and can be clicked. The scope of the issue is to disable the button if no template is selected.

You can check if the `template` state is set in the `NewModule` component. If not, the template is not selected, and the Save button should be disabled.

Abito12:
closes #503 

## 📑 Description
- Disabling save button in create new module until a template is selected

## ✅ Checks

### DIFF
--- a/cyclops-ui/src/components/pages/NewModule/NewModule.tsx
+++ b/cyclops-ui/src/components/pages/NewModule/NewModule.tsx
@@ -1298,7 +1298,7 @@ const NewModule = () => {
                 disabled={
                   loadingTemplate ||
                   loadingTemplateInitialValues ||
-                  !template.version
+                  !(template.version || template.path || template.repo)
                 }
               >
                 Save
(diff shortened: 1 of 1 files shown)


cyclops-ui/src/components/pages/NewModule/NewModule.tsx
```diff
@@ -1295,7 +1295,11 @@ const NewModule = () => {
                 }
                 htmlType="submit"
                 name="Save"
-                disabled={loadingTemplate || loadingTemplateInitialValues}
+                disabled={
+                  loadingTemplate ||
+                  loadingTemplateInitialValues ||
+                  !template.version
```
petar-cvit, line 1301: The version is not required for a template, so this might not work in some cases. Can you update the condition so that at least one of the `template.repo`, `template.path` or `template.version` needs to be set?

petar-cvit (changes requested):
Thanks @Abito12 for the PR. I left one comment

cyclops-ui/src/components/pages/NewModule/NewModule.tsx
abitoprakash: Updated the condition, please recheck.

petar-cvit (approved):
Thanks @Abito12 🧡 Tried it locally and it looks good

PR #517, opened 2024-08-21
fixed issue-344

### LINKED ISSUE #344: Expose modules errors on API endpoint
Each module renders K8s resources from a template. While doing that, some errors can occur. Those errors are already present on the CRD, but still need to be mapped to the DTO model.

You will need to add the errors field [here](https://github.com/cyclops-ui/cyclops/blob/aadf667833267684b5152418a332a558478179c1/cyclops-ctrl/internal/models/dto/modules.go#L3) and  map it in the [mapper function](https://github.com/cyclops-ui/cyclops/blob/aadf667833267684b5152418a332a558478179c1/cyclops-ctrl/internal/mapper/modules.go#L39).

patelajay745:
closes #344 

## ✅ Checks
- [X ] I have performed a self-review of my code

## ℹ Additional context
Please check and update me.


petar-cvit (approved):
Thanks @patelajay745  🧡

PR #521, opened 2024-08-22
Support multiple values module update

### LINKED ISSUE #519: Support multiple updates on `cyctl update module`
`cycle update module` currently supports updating a single value in the module. The scope of the task is to support updating multiple values at once.

The command is currently invoked like below:
```
cyctl update module my-app --key="scaling.replicas" --value=3
```

After supporting updating multiple values, each updated value should be defined with a single flag that defines both key and value:
```
cyctl update module my-app --value="scaling.replicas=3" --value="expose.enabled=true" 
```

rimvydascivilis:
closes #519 

## 📑 Description
Removed `key` flag and changed `value` flag to support multiple values.

## ✅ Checks

## ℹ Additional context
Used [StringArrayP](https://github.com/spf13/pflag/blob/v1.0.5/string_array.go#L127) [¶](https://pkg.go.dev/github.com/spf13/pflag?utm_source=godoc#StringArrayP) instead of [StringSliceP](https://github.com/spf13/pflag/blob/v1.0.5/string_slice.go#L161) [¶](https://pkg.go.dev/github.com/spf13/pflag?utm_source=godoc#StringSliceP), so `--ss="v1,v2" --ss="v3"` will not work.

### DIFF
--- a/cyctl/cmd/update.go
+++ b/cyctl/cmd/update.go
@@ -7,10 +7,10 @@ import (
 
 var (
 	updateExample = `# updates the given module 
-cyctl update module <module-name> --key=<key> --value=<value> 
+cyctl update module <module-name> --value="<key>=<value>" --value="<key>=<value>"
 
-# to update replicas for a module named test,updates number of replicas to 3
-cyctl update module <module-name> test --key="scaling.replicas" --value=3`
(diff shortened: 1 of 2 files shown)


petar-cvit (approved):
Thanks @rimvydascivilis 🧡

PR #546, opened 2024-09-01
🚀 feat: added delete pod functionality to the pod table

### LINKED ISSUE #538: Implement delete pod action

[PodTable](https://github.com/cyclops-ui/cyclops/blob/main/cyclops-ui/src/components/k8s-resources/common/PodTable/PodTable.tsx) showing pods for Deployments, DaemonSets and StatefulSets has a column called actions with a menu. When expanded, the menu currently only has a button for checking pod logs.

Screenshot 2024-08-27 at 14 44 11

The scope of the issue is to implement the delete functionality and add a button to the same menu. When the button is clicked there should be a confirm modal. When the deletion is confirmed, the UI should call the `/resources` backend endpoint with the delete method.

A similar button is implemented [here](https://github.com/cyclops-ui/cyclops/blob/af65091e12c2a44867013a39a7919586599aa433/cyclops-ui/src/components/pages/ModuleDetails.tsx#L1055).

naineel1209:
closes #538

## 📑 Description

- Added a new `Delete` button to the Pod Actions Menu

image

- Added a `Confirm Delete Modal` which confirms the deletion of pod and then makes a `DELETE` request to the backend endpoint `/resources`

image

- Added a new `triggerReload` state to the components that uses `PodTable` component, so that upon the deletion of the pod we can manually fetch the resources from the parent component.
- Added a new Prop `handleTriggerReloadUpdate` to the `PodTable`, which is used to call the trigger in the parent component.

https://github.com/user-attachments/assets/9084e919-0d45-4a95-9bf1-8e172d5d2aac

## ✅ Checks
- [✅] I have tested my code (provide screenshots or screen recordings of a working solution)

### DIFF
--- a/cyclops-ui/src/components/k8s-resources/common/PodTable/PodTable.tsx
+++ b/cyclops-ui/src/components/k8s-resources/common/PodTable/PodTable.tsx
@@ -28,6 +28,7 @@ import {
 interface Props {
   namespace: string;
   pods: any[];
+  handleTriggerReloadUpdate: () => void;
 }
 
 interface Pod {
@@ -38,7 +39,7 @@ interface Pod {
   namespace: any;
 }
 
-const PodTable = ({ pods, namespace }: Props) => {
+const PodTable = ({ pods, namespace, handleTriggerReloadUpdate }: Props) => {
   const [logs, setLogs] = useState("");
   const [logsModal, setLogsModal] = useState({
     on: false,
@@ -103,6 +104,8 @@ const PodTable = ({ pods, namespace }: Props) => {
           namespace: deletePodRef.podDetails.namespace,
         },
       });
+
+      handleTriggerReloadUpdate();
 
       setDeletePodConfirmRef("");
       setDeletePodRef({
@@ -450,7 +453,7 @@ const PodTable = ({ pods, namespace }: Props) => {
             Delete
           </Button>
         }
-        width={"30%"}
+        width={"40%"}
         styles={{
           footer: {
             display: "flex",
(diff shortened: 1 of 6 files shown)


cyclops-ui/src/components/k8s-resources/common/PodTable/PodTable.tsx
```diff
@@ -160,49 +248,78 @@ const PodTable = ({ pods, namespace }: Props) => {
...
+                initContainers: pod.initContainers,
+              });
+            }}
+          >
+            <h4>
+              <ReadOutlined style={{ paddingRight: "5px" }} />
+              View Logs
+            </h4>
+          </Button>
+          <Button
```
petar-cvit, line 302: Can we make the buttons go one under the other? There will be a third button for pod manifest, and it will probably look better if they are one under the other as opposed to being in the same line.

petar-cvit (changes requested):
@naineel1209 thanks for the PR!

I get where you are going with `triggerReload` state, but I have another idea that will be easier to maintain.

I would prefer if `PodTable` component receives a callback function that will be ran when a pod is deleted. Then, each parent component (Deployment, StatefulSet...) will pass a function that fetches the latest state of the resource and updates its state.

I'm suggesting it because there are no additional states to the components this way

cyclops-ui/src/components/k8s-resources/common/PodTable/PodTable.tsx
naineel1209: resolved in e6dfca2

naineel1209:
@petar-cvit As per your suggestions, I've updated the code to use a callback to update the data after the pod is deleted.

petar-cvit (approved):
Thanks @naineel1209 🧡

PR #568, opened 2024-09-12
🚀 feat: implement logs streaming for pods

### LINKED ISSUE #555: Stream pod logs
Currently, Cyclops UI supports fetching logs of a pod, but logs are not streamed in real time. The scope of the issue is to implement log streaming for pods on both the backend and frontend.

You can use SSE to stream updates. https://github.com/cyclops-ui/cyclops/pull/544 implements streaming for resources so similar libs can be used.

naineel1209:
closes #555 

## 📑 Description
[BE]
- Implemented streaming logs of pod continuously to the client using SSE.
- Utilized previously coded SSE implementation and made channels to stream the logs to the FE.

[FE]
- Changed `logs` state to store array of strings to easily manage logs by using `logs.join("\n")`.
- Implemented `logStream` function which streams the logs and sets it in the state.
- Handled scenario when no logs are available, container tabs are changed and when the log modal is cancelled or OKed.
- Introduced a new `logsSignalController` state to be passed to the `logStream` function to handle the SSE connection scenarios i.e. close the connection via signal when modal is closed, close the current connection via signal and make a new signal for the new connection when the container tab is changed.

Streamed Logs Demo + No more events fired after the modal is closed:
https://github.com/user-attachments/assets/1f3f5fdf-102d-4697-9188-002092279098

## ✅ Checks
- [✅] I have tested my code (provide screenshots or screen recordings of a working solution)


cyclops-ui/src/utils/api/sse/resources.tsx
```diff
@@ -54,3 +54,60 @@ export function resourceStream(
     },
   }).catch((r) => console.error(r));
 }
+
+export function logStream(
```
petar-cvit, line 58: Can you move this to a new file `logs.tsx`

cyclops-ctrl/pkg/cluster/k8sclient/client.go
```diff
@@ -169,6 +171,34 @@ func (k *KubernetesClient) GetPods(namespace, name string) ([]apiv1.Pod, error)
 	return podList.Items, err
 }
 
+func (k *KubernetesClient) GetStreamedPodLogs(ctx *gin.Context, namespace, container, name string, logCount *int64, logChan chan<- string) error {
```
petar-cvit, line 174: Can you pass only the context of `ctx`. I don't want to import gin stuff into k8s client since those are not connected in any way.

cyclops-ui/yarn.lock
```diff
@@ -5278,9 +5278,9 @@ caniuse-api@^3.0.0:
     lodash.uniq "^4.5.0"
 
 caniuse-lite@^1.0.0, caniuse-lite@^1.0.30001587, caniuse-lite@^1.0.30001591:
-  version "1.0.30001597"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001597.tgz#8be94a8c1d679de23b22fbd944232aa1321639e6"
-  integrity sha512-7LjJvmQU6Sj7bL0j5b5WY/3n7utXUJvAe1lxhsHDbLmwX9mdL86Yjtr+5SRCyf8qME4M7pU2hswj0FpyBVCv9w==
+  version "1.0.30001658"
+  resolved "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001658.tgz"
+  integrity sha512-N2YVqWbJELVdrnsW5p+apoQyYt51aBMSsBZki1XZEfeBCexcM/sf4xiAHcXQBkuOwJBXtWF7aW1sYX6tKebPHw==
```
petar-cvit, line 5283: Did we need to bump these? If not, please revert it

petar-cvit (changes requested):
@naineel1209 thanks for the PR! Looks awesome 🧡 

Left some comments on code structure

cyclops-ui/yarn.lock
naineel1209: While building the `cyclops-ui`, it was showing this warning 

image

but since this bump doesn't affect the build, I've reverted it.

cyclops-ui/src/components/k8s-resources/common/PodTable/PodLogs.tsx
```diff
@@ -99,31 +114,32 @@ const PodLogs = ({ pod }: PodLogsProps) => {
...
-            log += s;
-            log += "\n";
+    setLogsSignalController((prevController) => {
+      if (prevController) {
+        prevController.abort();
+      }
+
+      const controller = new AbortController();
+      return controller;
+    });
```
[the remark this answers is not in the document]
naineel1209: Thanks for highlighting this issue. It might arise in some cases, so to completely avoid the situation, we can add the following code (which I used in the file `PodLogs.tsx`, lines 164–180):

```typescript
const controller = new AbortController();
setLogsSignalController(() => controller);

logStream(
  pod.name,
  pod.namespace,
  pod.containers[0].name,
  (log) => {
    setLogs((prevLogs) => [...prevLogs, log]);
  },
  (err) => {
    setError(mapResponseError(err));
  },
  controller,
);
```

We handle the required modification by aborting the previous signal controller, and then creating and passing a **new signal controller** from the closure. This ensures that the new `AbortController` is always passed to the `logStream` function, making sure an active abort signal controller is present in the `logStream`.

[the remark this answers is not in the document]
petar-cvit: Sure, feel free to add it to the PR

petar-cvit (changes requested):
@naineel1209, I tried testing it and noticed that when I fetch the logs, those are streamed as expected, but when the window with my Cyclops is not in focus or I change the tab in my browser, it stops the connection and refetches all the logs again. This results in duplicated logs.

Could Abort controller be at fault? To be honest if its the root cause we can remove it, but could you look into it?

naineel1209:
Hi @petar-cvit, I examined the issue, and it was happening due to the default behaviour of the @microsoft/fetch-event-source library, which closes the SSE events when the document is not visible. This caused our request to close as soon as the tab's visibility changed and restart when the tab was in focus again, resulting in fetching the latest 100 logs and streaming the new logs while our old logs were already present.

I have resolved this issue by setting the openWhenHidden flag to true, which keeps the SSE connection open when the tab's visibility changes.

petar-cvit:
@naineel1209, that makes sense. This didn't happen when getting updates for workloads getting a new update would not affect data.

However, if the sse gets disconnected for some other reason (e.g. load balancer having a timeout of one minute) it will still refetch all the logs and append them.

Can you handle the case when see reconnects? We should still keep the `openWhenHidden`, but we need to handle the case of the sse disconnect

naineel1209:
@petar-cvit, I reviewed the disconnection issue that leads to duplicate logs upon reconnection and an unresolved error modal, even after a successful reconnection. To tackle this, I introduced an `isReset` flag, which ensures that both the logs and the error modal are properly reset upon successful connection and during future reconnections, allowing for a fresh stream of logs.

naineel1209:
Hi @petar-cvit, any updates on this?

petar-cvit (approved):
Thanks for the PR @naineel1209! I [removed the abort controller state](https://github.com/cyclops-ui/cyclops/pull/568/commits/9d477011401c1f1e708ccc8b1a1b184625053ca0) since it wasn't used.

Also, we need to cover the case when streaming is disabled, but I will do it in a separate PR

PR #620, opened 2024-10-08
feat(cyclops-ui): Dropdown menu to accept other then just enum fields

### LINKED ISSUE #50: Dropdown menu to accept other then just enum fields
In some cases, a dropdown is useful just as a suggestion, not as only options for the field

Add a new field `suggestions` that behaves the same way as `enum`, but the UI will allow you to add any text you decide on

s-vamshi:
closes #50

## 📑 Description

Dropdown menu to accept other then just enum fields when json key has x-suggestions

## ✅ Checks

## ℹ Additional context
[screen-capture (15).webm](https://github.com/user-attachments/assets/815fbacd-644f-4819-a37b-b133607b00cc)


s-vamshi:
hey @petar-cvit can you please review this pr in your free time?

s-vamshi:
@quest-bot loot #50

KaradzaJuraj (changes requested):
Hey @s-vamshi, I’m having trouble running this code. It seems that `mode: ”tags"` sends the values as arrays, which breaks on the backend. I would much rather prefer a dropdown that sends a single value instead of an array of one value. 

Check an example of `Select` with a [Custom dropdown](https://ant.design/components/select#select-demo-custom-dropdown-menu) from AntDesign to get an idea of how you could implement this.

If it makes sense, you can extract it into a separate component (something like `SuggestInput.tsx` perhaps).

Let me know if you can make it work and if you encounter issues with this approach!

s-vamshi:
Hey @KaradzaJuraj, thanks for the suggestions, I will use custom drop-down in a new component instead of making changes to SelectInput component!

s-vamshi:
Hey @KaradzaJuraj @petar-cvit added a new suggestionInput field using custom dropdown
image

[demo](https://github.com/user-attachments/assets/9dadf4e2-0920-44ee-a8ce-b1c1c4b031fd)

cyclops-ui/src/components/form/fields/string/SuggestionInput.tsx
```diff
@@ -0,0 +1,133 @@
...
+            {menu}
+            <Divider style={{ margin: "8px 0" }} />
+            <Row>
+              <Col span={16}>
+                <Input
+                  placeholder="Enter New Option"
+                  value={newOption}
+                  onChange={(event) => setNewOption(event.target.value)}
+                  onKeyDown={(e) => e.stopPropagation()}
+                  onKeyPress={(e) => e.key === "Enter" && addOptionOnEnter()}
```
KaradzaJuraj, line 115: `OnKeyPress` is deprecated; can you move the functionality to `onKeyDown`?

KaradzaJuraj (changes requested):
Looks great @s-vamshi! Just one more change and we can add this to the codebase 🚀

cyclops-ui/src/components/form/fields/string/SuggestionInput.tsx
s-vamshi: hey @KaradzaJuraj, moved the logic to onkeydown!

s-vamshi:
hey @KaradzaJuraj done with the changes!

KaradzaJuraj (approved):
Awesome! Thanks for picking this up @s-vamshi 🙏

PR #665, opened 2024-10-31
Add ClusterRole resource

### LINKED ISSUE #436: ClusterRole Backend Implementation

```bash
```

The goal of the issue is to support ClusterRole on the same endpoint. The endpoint for fetching ClusterRole data would look like this: 

```bash
 /resources?group=apps&version=v1&kind=ClusterRole&name=&namespace=
```

The endpoint should return data about its `name`, `rules`, `resources`, `verbs`… (you can find more about it [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#clusterrole-v1-rbac-authorization-k8s-io)

nikonhub:
closes #436

## 📑 Description
Add a clusterRole kind information to resource endpoint

## ✅ Checks

## ℹ Additional context
image

### DIFF
--- a/cyclops-ctrl/internal/models/dto/k8s.go
+++ b/cyclops-ctrl/internal/models/dto/k8s.go
@@ -709,3 +709,45 @@ func (s *Role) GetDeleted() bool {
 func (s *Role) SetDeleted(deleted bool) {
 	s.Deleted = deleted
 }
+
+type ClusterRole struct {
+	Group     string              `json:"group"`
+	Version   string              `json:"version"`
+	Kind      string              `json:"kind"`
+	Name      string              `json:"name"`
+	Namespace string              `json:"namespace"`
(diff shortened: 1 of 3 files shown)


nikonhub:
@petar-cvit,

I added clusterRole kind to resource endpoint.

I have a few questions:
- Is it outside the scope of this issue to integrate this with the front end?
- A ClusterRole has no namespace. I believe it should still implement the Resource interface (GetNamespace method). Should we just let an empty string ?

petar-cvit (approved):
Looks good! Thanks 🧡

petar-cvit:
@nikonhub If you are up for it, you can implement the frontend part in a separate PR. And yeah, Cluster Role is non-namespaced, but it still implements the interface as you did it here.

You can turn the PR from draft to ready to review and I can merge it 😄

petar-cvit (approved):

PR #668, opened 2024-11-01
Add cluster role section to a module

### LINKED ISSUE #436: ClusterRole Backend Implementation

nikonhub:
closes #436

## 📑 Description
Add ClusterRole section to module resources

## ✅ Checks

## ℹ Additional context
image


cyclops-ui/src/components/k8s-resources/ClusterRole.tsx
```diff
@@ -0,0 +1,148 @@
...
+      title: "API Groups",
+      dataIndex: "apiGroups",
+      key: "apiGroups",
+      render: (apiGroups: string[]) => (
+        <>
+          {apiGroups.map((group) => (
+            <Tag key={group} color="blue">
+              {group || "*"}
+            </Tag>
+          ))}
```
petar-cvit, line 75: Not all rules will have `apiGroups`. With cluster roles, you can have a rule that has a `nonResourceURLs` field, and those rules will not have the `apiGroups` property.

You can create a Module from the Prometheus bitnami helm chart and you will get one such cluster role which  has a single rule without `apiGroups`.

Because of that, the `apiGroups` here will be null and calling `.map` will break the UI

petar-cvit (changes requested):
@nikonhub thanks for the PR, looks awesome! I left one comment on one edge case.

cyclops-ui/src/components/k8s-resources/ClusterRole.tsx
nikonhub: Yes right. I've fixed this case.

image

Should we display something for this rule ?

```json
{
	"2": {
		"verbs": [
			"get"
		],
		"nonResourceURLs": [
			"/metrics"
		]
	}
}
```

petar-cvit: It might be a bit confusing if you see a rule without any resources and just the verbs.
Can we conditionally add a column to this table? I was thinking of adding a nonResourceURLs column if there is at least one rule with the property. If not, I would leave those three columns. In that case, I would put the verbs as the first column, followed by APIGroups and resources and, if needed, nonREsourceURLs.

petar-cvit: @nikonhub, can you check my comment above and add a column for nonResourceURLs if possible

PR #889, opened 2026-05-12
Fix resource manifest modal loading state

### LINKED ISSUE #641: Manifests modals to use the same theme
`Module Manifest` and `Rendered Manifest`, use a different modal compared to the `View Manifest` button that every resource has. `View Manifest` needs to be updated.

Create a new state `loadingResourceManifest`, which will be set to `true` while fetching the resource manifest. Once fetched (successful or not), set it to `false` (the same way how the `loadingRawManifest` and `loadingRenderedManifest` states work).

Then use the `moduleManifestContent()` function to render the manifest (check how the Rendered manifest Modal does it).

Additionally, remove the `OK` button.

anonymusk7:
## Summary
- add a loading state for resource manifest fetches
- render resource manifests with the same Ace-based layout/options used by the module/rendered manifest modals
- remove the resource manifest modal OK footer so it closes via the modal close/cancel affordance

Closes #641.

## Validation
- `npx --yes prettier@3.2.5 --check cyclops-ui/src/components/k8s-resources/ResourceList/ResourceList.tsx`
- `git diff --check`

## Note
I attempted `yarn --cwd cyclops-ui install --frozen-lockfile --ignore-scripts` twice, including with `--network-timeout 600000`, but dependency fetching stalled in this environment before `node_modules` was created. I did not claim a full local build.


anonymusk7:
@quest-bot loot #641

PR #890, opened 2026-05-13
Support auth for OCI helm templates

### LINKED ISSUE #524: Support OCI helm registry auth
When fetching private templates, Cyclops injects auth data (username and password) into the request. This is supported only when fetching templates from GitHub, but not when fetching from an OCI registry.

[Here](https://github.com/cyclops-ui/cyclops/blob/ab9f84e3e264aca7bd5bf678d3c5c53f7f25c063/cyclops-ctrl/internal/template/git.go#L33) is the example of how credentials are fetched from the cluster when fetching a template from GitHub and [here](https://github.com/cyclops-ui/cyclops/blob/02fb7f027380d8cf577a5b13891142db72e24e5e/cyclops-ctrl/internal/template/oci.go#L16) is the function fetching data from an OCI based registry.

anonymusk7:
## Summary
- Resolve `TemplateAuthRule` credentials for OCI template repositories, matching the existing private Git template flow.
- Thread those credentials through OCI tag, token, manifest, and blob requests so private registries can authenticate with Basic auth or registry-issued bearer tokens.
- Keep the existing public `GetOCIChartTags` API as a no-credentials wrapper and add focused auth/token tests.

Fixes #524

## Tests
- `go test ./pkg/auth ./pkg/template ./internal/controller/tests`
- `git diff --check`

@quest-bot loot #524


anonymusk7:
@quest-bot loot #524

